PR #79, opened 2019-03-08
Add Fault Tolerance quickstart

TomasHofman:
Quickstart for MP Fault Tolerance.

There is also related docs PR: https://github.com/quarkusio/quarkus/pull/1318


gsmet:
Thanks @TomasHofman .

Could you also add it to the README? Just below the OpenTracing one would be perfect.

PR #81, opened 2019-03-08
add .s2i/environment to getting-started for OOB OpenShift S2I support

vorburger:


vorburger:
NB: This is with `<uberJar>true`, because that's what works best as-is today.

An alternative without that is proposed for discussion in #82.

getting-started-openshift-s2i/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                    </systemProperties>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-maven-plugin</artifactId>
+                <version>${quarkus.version}</version>
+                <configuration>
+                    <uberJar>true</uberJar>
```
cescoffier: Didn't we find a solution to avoid this?

README.md
```diff
@@ -7,7 +7,8 @@ See [CONTRIBUTING](CONTRIBUTING.md) for how to build these examples:
 * [Getting Started](./getting-started): Application creation, Rest endpoint, Dependency Injection, Test, Packaging
 * [Getting Started - Async](./getting-started-async): Illustrate how to use `CompletionStage` to handle asynchronous actions
 * [Getting Started - Native](./getting-started-native): Packaging of the application into a native executable
-* [Getting Started - Kubernetes](./getting-started-kubernetes): Deployment of the application to Kubernetes and/or OpenShift
+* [Getting Startied -OpenShift S2I](./getting-started-openshift-s2i): Deployment of the application to OpenShift with S2I
+* [Getting Started - Kubernetes/OpenShift via Dockerfile](./getting-started-kubernetes): Deployment of the application to Kubernetes and/or OpenShift
```
cescoffier: I would put the S2I approach after the Kubernetes guide. 
Also, a space is missing.

getting-started-openshift-s2i/pom.xml
vorburger: see #82

vorburger:
#87 will cause a (minor) conflict with this PR; ping me to rebase and fix that when #87 is merged.

vorburger:
@cescoffier what do we want to do re. this one?

vorburger:
@cescoffier and I discussed yesterday that he really much prefers having #82 folded into this one, so I've cherry-picked 002fc1473cc33a0483b665f849dda7bd3e48d395 from there to here. We also concluded that we do NOT want a separate `getting-started-openshift-s2i` but instead will just add the (required..) `.s2i/environment` to `getting-started`. I'll therefore move it according in this PR, and re-test it all together. 

FYI @tqvarnst @siamaksade @geoand @kameshsampath

getting-started/.s2i/environment
```diff
@@ -0,0 +1,6 @@
+JAVA_APP_JAR=*-runner.jar
```
vorburger: NOK, this doesn't actually work (with `registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift`)

vorburger:
@cescoffier @gastaldi @geoand this is ready for your review and merge from my side now.

getting-started/.s2i/environment
geoand: Very interesting to know, :+1:

geoand:
Looks good to me :+1:

getting-started/.s2i/environment
cescoffier: :-(

getting-started/pom.xml
```diff
@@ -48,6 +48,8 @@
     </dependencies>
 
     <build>
+        <!-- required because we can't use a *-runner.jar in .s2i/environment's JAVA_APP_JAR -->
+        <finalName>${project.artifactId}</finalName>
```
cescoffier: It creates an issue with the getting started guide that does not use this, and so it contains the path to the "long" name. Can we use the "long" name in the `environment` file?

cescoffier:
We would need some words in the Kubernetes - OpenShift guide (main quarkus repository)

getting-started/.s2i/environment
kameshsampath: in my builder I used something like this to compute the actual  build artifact name

```
ARTIFACT_NAME=$(mvn org.apache.maven.plugins:maven-help-plugin:3.1.1:evaluate -Dexpression=project.build.finalName -q -DforceStdout)
ARTIFACT_NAME_PKG=$(mvn org.apache.maven.plugins:maven-help-plugin:3.1.1:evaluate -Dexpression=project.packaging -q -DforceStdout)

# compute the app name with packaging
APP_NAME="$ARTIFACT_NAME.$ARTIFACT_NAME_PKG"
```


https://github.com/redhat-developer-demos/quarkus-java-builder/blob/master/bin/maven-buildah.sh#L15-L19

getting-started/pom.xml
kameshsampath: @cescoffier  - check my comment above a technique I followed in my builder image to compute the actual artifactName.

Not sure how much that will help in this context but just a thought

cescoffier: Yes, because it requires buildah right?

kameshsampath: no. i just used in a script to find the final artifact name.. it can be used with any context vanilla docker, buildah, kaniko etc

vorburger:
> We would need some words in the Kubernetes - OpenShift guide (main quarkus repository)

yep, that's proposed over in #1353 ...

getting-started/pom.xml
vorburger: > It creates an issue with the getting started guide that does not use this

You mean we have we hard-coded somewhere else (where?) that we expect the built JAR to have a version? That won't work because we can't use a *-runner.jar in .s2i/environment's `JAVA_APP_JAR` ... then we either have to find another way (I briefly played with `S2I_SOURCE_DEPLOYMENTS_FILTER` in #90 [before squashing](https://github.com/quarkusio/quarkus-quickstarts/compare/master...vorburger:issue-1345_s2i-HISTORY-BEFORE-SQUASH), but it didn't work - perhaps that's worth another attempt...), or work on a patch to https://github.com/jboss-container-images/openjdk ... phew.

vorburger: I think I've figured out a solution for this (see 9adaa4b22bb4a1a8ca785329a2799ecc665211fc)... just need to fully re-test it e2e ...

getting-started/.s2i/environment
vorburger: Resolving this, as I've now been able to figure out the right magic combination which does the trick to make this work.

vorburger:
@cescoffier please re-review, I expect you'll like this better now... :small_airplane: 

@gastaldi check out how this avoids fixing a <build><finalName> now.

getting-started/pom.xml
vorburger: Resolving this conversation, as 9adaa4b indeed does the trick and works great.

getting-started/.s2i/environment
```diff
@@ -0,0 +1,5 @@
+# ARTIFACT_COPY_ARGS is deprecated, but S2I_SOURCE_DEPLOYMENTS_FILTER does not actually work, see https://github.com/jboss-container-images/openjdk/issues/75
+
+ARTIFACT_COPY_ARGS=-p -r lib/ *-runner.jar
+
+JAVA_OPTIONS=-Dquarkus.http.host=0.0.0.0
```
gastaldi: Better add the options to avoid creating the quarkus.log file too?

geoand: It shouldn't be necessary with: https://github.com/quarkusio/quarkus/pull/1603

cescoffier:
Does it work with the latest release or does it required SNAPSHOT?

As this PR is open against master, it has to tun against the latest release. But I guess it's not the case. So we would need to re-open it against the _development_ branch.

vorburger:
> Does it work with the latest release or does it required SNAPSHOT?

it works just fine against the latest release as-is! (I tested it.)

> As this PR is open against master, it has to tun against the latest release. 

Given above, is it OK to keep it for `master` then? Merge?

> So we would need to re-open it against the _development_ branch.

FYI GitHub Trick: PR Edit (upper right hand corner) let's you change the _into_ branch - useful.

PR #82, opened 2019-03-08
[WIP] avoid using <uberJar>true in getting-started-openshift-s2i

vorburger:


vorburger:
This is WIP, because IMHO we should not merge this until we have https://github.com/quarkusio/quarkus/pull/1345 from @geoand merged and released; then we can clean this one up (avoid the `finalName` in the `pom.xml` and the `JAVA_APP_JAR` in `.s2i/environment`), retest it, and then get it in.

In the mean time, I would suggest that we already merge #81, for now.

vorburger:
Closing this as it will be worked into #81.

Issue #70, opened 2019-03-07
TokenSecuredResource#helloShouldDeny(SecurityContext) has wrong annotation


starksm64:
One of the solution method has an incorrect `@PermitAll` annotation that should have been `@DenyAll`:

```java
    @GET()
    @Path("deny-all")
    @PermitAll
    @Produces(MediaType.TEXT_PLAIN)
    public String helloShouldDeny(@Context SecurityContext ctx) {
        Principal caller =  ctx.getUserPrincipal();
        String name = caller == null ? "anonymous" : caller.getName();
        return "hello + "+name;
    }
```
Should be:
```java
    @GET()
    @Path("deny-all")
    @DenyAll
    @Produces(MediaType.TEXT_PLAIN)
    public String helloShouldDeny(@Context SecurityContext ctx) {
        Principal caller =  ctx.getUserPrincipal();
        String name = caller == null ? "anonymous" : caller.getName();
        return "hello + "+name;
    }
```

The associated `TokenSecuredResourceTest#testHelloDenyAll` needs to be updated to expect a HttpURLConnection.HTTP_FORBIDDEN rather than HttpURLConnection.HTTP_OK.

cescoffier:
Fixed in master.

PR #83, opened 2019-03-09
Fix the #70 issue

starksm64:
Add workaround for quarkus issue 1292


cescoffier:
@starksm64 just do be sure, this fix is for 0.11.0 right (so last release).

starksm64:
@cescoffier Yes, that is correct.

cescoffier:
Thanks @starksm64, it's merged!

PR #87, opened 2019-03-10
Merge the native and kubernetes quickstart in getting started

cescoffier:
Because the project generated by the user contains everything required, no need for 3 quickstarts anymore. 

Also, this PR make the instructions more consistent.


getting-started/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -0,0 +1,2 @@
+# Configuration file
+key = value
```
gsmet: This file should be removed.

cescoffier: File deleted.

cescoffier:
Yes, this PR is against `master` on purpose. It's an improvement of the guide that should be publish soon (without a release)

PR #91, opened 2019-03-11
Add .dockerignore to make the context to send lighter

anthonydahanne:


cescoffier:
That’s a great idea. I believe we should generate it with the project. WDYT?

anthonydahanne:
yes, definitely; the generator that creates the ```Dockerfile``` should also generate the ```.dockerignore```

closing this PR in favor of this one : https://github.com/quarkusio/quarkus/pull/1407

PR #96, opened 2019-03-13
improved user-experience in chat example

domdorn:
improved the user experience for the chat example:
- when loading the page, focus will jump to the name field
- when pressing enter in the name field, websocket gets connected, focus jumps to message field
- when pressing enter in the message field, message gets sent
- scrolling to the last text now works when lots of messages are posted


gsmet:
Nice, thanks!

PR #100, opened 2019-03-14
fix repository url from jbossas to quarkusio

kdnakt:


gsmet:
Good catch, thanks for the PR.

PR #102, opened 2019-03-15
feature(getting-started): Added DMP for building the image

rhuss:
Advantages:

* Less context to be send to the Docker daemon (because of .maven-dockerinclude)
* Image name can be configured in the pom.xml so that a plain "mvn docker:build" is enough.

Moved up also the Dockerfile to make it more prominent and aligned to other quickstarts.
This is not necessarily needed to use dmp


getting-started/Dockerfile
```diff
@@ -5,7 +5,12 @@
 #
 # Then, build the image with:
 #
-# docker build -f src/main/docker/Dockerfile -t quarkus/getting-started .
+# mvn docker:build
```
cescoffier: For my information, does it pick this docker file automatically?

I would also keep both with some explanation, so the user can decide and understand what does the Docker Maven Plugin.

getting-started/pom.xml
```diff
@@ -71,6 +74,12 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <plugin>
+                <groupId>io.fabric8</groupId>
+                <artifactId>docker-maven-plugin</artifactId>
+                <version>0.28.0</version>
```
cescoffier: Can we extract this version to a variable?

rhuss: Yes, sure.

getting-started/Dockerfile
rhuss: Yes. If not configured otherwise, d-m-p looks for a Dockerfile in the top-level directory. You can also configure to use a Dockerfile in a different place and e.g. also reuse the Dockerfile in src/. 

Also, you have Maven property interpolation within the Dockerfile (but then it can't be reused with plain `docker`, or only with some tricks using buildArgs, which I wouldn't recommend).

A use case for property interpolation would be to add meta data to the image automatically, like the one defined in http://label-schema.org/rc1/ (like git ref etc.). Of course you could achieve the same with the resource plugin, but for dmp this is more direct.

If you like I can add such labels to the PR, too (after Javaland is over, though).

rhuss:
I am closing it here since the PR is already 5 years old, and I lost the context :-)

PR #108, opened 2019-03-18
[DO NOT MERGE] Declare the test scope on test dependencies

cescoffier:
Follow up from https://github.com/quarkusio/quarkus/pull/1557


gsmet:
@cescoffier I merged it as it can be merged now. We don't need the main repo PR for it to work as we add more information.

PR #130, opened 2019-04-02
Reactive Postgres Client quickstart

tsegismont:
A sample RESTful application using the Reactive Pg Client to manage fruit entities.

### DIFF
--- a/using-vertx/src/main/java/org/acme/vertx/FruitResource.java
+++ b/using-vertx/src/main/java/org/acme/vertx/FruitResource.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CompletionStage;
 import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.ResponseBuilder;
 import javax.ws.rs.core.Response.Status;
@@ -32,8 +33,8 @@ import org.jboss.resteasy.annotations.jaxrs.PathParam;
 import io.reactiverse.axle.pgclient.PgPool;
 
 @Path("fruits")
-@Produces("application/json")
-@Consumes("application/json")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
 public class FruitResource {
 
     @Inject
--- a/using-vertx/src/main/java/org/acme/vertx/FruitResource.java
+++ b/using-vertx/src/main/java/org/acme/vertx/FruitResource.java
@@ -63,7 +63,7 @@ public class FruitResource {
     @GET
     public CompletionStage<Response> get() {
         return Fruit.findAll(client)
-                .thenApply(fruits -> fruits.isEmpty() ? Response.noContent() : Response.ok(fruits))
+                .thenApply(Response::ok)
                 .thenApply(ResponseBuilder::build);
     }
 
--- a/using-vertx/src/main/java/org/acme/vertx/FruitResource.java
+++ b/using-vertx/src/main/java/org/acme/vertx/FruitResource.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.acme.vertx;
+
+import java.net.URI;
+import java.util.concurrent.CompletionStage;
+
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+import javax.ws.rs.*;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.ResponseBuilder;
+import javax.ws.rs.core.Response.Status;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.resteasy.annotations.jaxrs.PathParam;
+
+import io.reactiverse.axle.pgclient.PgPool;
+
+@Path("fruits")
+@Produces("application/json")
+@Consumes("application/json")
+public class FruitResource {
+
+    @Inject
+    @ConfigProperty(name = "myapp.schema.create", defaultValue = "true")
+    boolean schemaCreate;
+
+    @Inject
+    PgPool client;
+
+    @PostConstruct
+    void config() {
+        if (schemaCreate) {
+            initdb();
+        }
+    }
+
+    private void initdb() {
+        client.query("DROP TABLE IF EXISTS fruits")
+                .thenCompose(r -> client.query("CREATE TABLE fruits (id SERIAL PRIMARY KEY, name TEXT NOT NULL)"))
+                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Orange')"))
+                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Pear')"))
+                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Apple')"))
+                .toCompletableFuture()
+                .join();
+    }
+
+    @GET
+    public CompletionStage<Response> get() {
+        return Fruit.findAll(client)
+                .thenApply(fruits -> fruits.isEmpty() ? Response.noContent() : Response.ok(fruits))
+                .thenApply(ResponseBuilder::build);
+    }
+
+    @GET
+    @Path("{id}")
+    public CompletionStage<Response> getSingle(@PathParam Long id) {
+        return Fruit.findById(client, id)
+                .thenApply(fruit -> fruit != null ? Response.ok(fruit) : Response.status(Status.NOT_FOUND))
+                .thenApply(ResponseBuilder::build);
+    }
+
+    @POST
+    public CompletionStage<Response> create(Fruit fruit) {
+        return fruit.save(client)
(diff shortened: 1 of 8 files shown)


using-vertx/src/main/java/org/acme/vertx/FruitResource.java
```diff
@@ -0,0 +1,98 @@
...
+        });
+        CompletionStage<Response> response = fruits
+                .thenApply(list -> list.isEmpty() ? Response.noContent() : Response.ok(list))
+                .thenApply(ResponseBuilder::build);
+        return response;
+    }
+
+    @GET
+    @Path("{id}")
+    public CompletionStage<Response> getSingle(@PathParam("id") Long id) {
```
FroMage: you're using the right annotation: if you configure your maven compiler to include method parameter names you can just use `@PathParam` with no parameter.

using-vertx/src/main/java/org/acme/vertx/FruitResource.java
```diff
@@ -0,0 +1,98 @@
...
+        return client.preparedQuery("SELECT id, name FROM fruits WHERE id = $1", Tuple.of(id))
+                .thenApply(PgRowSet::iterator)
+                .thenApply(iterator -> iterator.hasNext() ? toFruit(iterator.next()) : null)
+                .thenApply(fruit -> fruit != null ? Response.ok(fruit) : Response.status(Status.NOT_FOUND))
+                .thenApply(ResponseBuilder::build);
+    }
+
+    @POST
+    public CompletionStage<Response> create(Fruit fruit) {
+        return client.preparedQuery("INSERT INTO fruits (name) VALUES ($1) RETURNING (id)", Tuple.of(fruit.name))
```
FroMage: postgres DOES THAT??? I never knew :(

tsegismont: Well, not hiding SQL must have some upsides :wink: 

Search for "modern SQL" presentations on the web, there are many many thing Postgres (and other RDBs) can do.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Inject
+    Vertx vertx;
+
+    @Produces
+    PgPool client;
+
+    @PostConstruct
+    void config() {
+        PgPoolOptions options = new PgPoolOptions()
```
cescoffier: I would have prefered an extension producing the client from the configuration instead of letting the user creating the instance.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
```diff
@@ -0,0 +1,85 @@
...
+
+        client = PgClient.pool(vertx, options);
+
+        if (schemaCreate) {
+            initdb();
+        }
+    }
+
+    private void initdb() {
+        client.query("DROP TABLE IF EXISTS fruits")
```
cescoffier: Can't it reuse the import.sql support?

using-vertx/src/main/resources/application.properties
```diff
@@ -1 +1,5 @@
-quarkus.ssl.native=true
\ No newline at end of file
+quarkus.ssl.native=true
```
cescoffier: any reason for SSL?

cescoffier:
I'm a bit puzzled by the different experience it's providing in comparison to Panache. I believe some integration work is required.

using-vertx/src/main/java/org/acme/vertx/FruitResource.java
tsegismont: But then we need to tell the user to do this in the guide, don't we? For people following the guide and just looking at the quickstart it will be fine though.

using-vertx/src/main/resources/application.properties
FroMage: I think it's due to Postgres connecting by default over SSL.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
FroMage: Yeah, Panache RX is going to try to offer that too. ATM it's ORM that's doing the import, so there's nothing he can use to do that.

using-vertx/src/main/java/org/acme/vertx/FruitResource.java
FroMage: We should tell them, yes.

tsegismont:
>We already have a demo for the reactive-pg-client (using rxjava not axle) at https://github.com/jbossas/protean-rest-http-crud-demo/tree/master/quarkus-vertx/src/main/java/io/quarkus

GH says not found when I click on this link.

>I think we should change our demo to make it use axle, but I think it's cleaner if we keep the DB operations inside the Fruit class like I did. WDYT?

I don't think about the `Fruit` class as a domain object (it's more a transfer object as a matter of fact). But I have no preference. If you want to keep the style consistent it's fine with me.

using-vertx/src/main/resources/application.properties
tsegismont: I can't remember why exactly. I believe you added it when you participated in the "Using Vert.x" quickstart PR :smile_cat:

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
tsegismont: We could but for what benefit? It would add complexity to this method which is used as a snippet in the guide to introduce the `query` client operation.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
tsegismont: This would be ideal. Can you file an issue and assign to me? I can work on it later and update the guide.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
FroMage: No, I think what @cescoffier meant was that if you had a file named `import.sql` in your resource folder, then quarkus should take care of doing that insertion for you.

FroMage:
I've asked for permission to be granted to you to look at it. Then we can debate which one looks better. My guess is that we may end up with a mix of the two.

tsegismont:
>I'm a bit puzzled by the different experience it's providing in comparison to Panache. I believe some integration work is required.

It's not perfect. As you said, an extension for the Reactive Pg Client would be an improvement. However the configuration bean uses Quarkus tech (Arc, Microprofile) so it's not terrible either.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
tsegismont: @FroMage there's no `import.sql` in this quickstart for now. Tell me what's expected because your last comment confused me.

tsegismont:
>I've asked for permission to be granted to you to look at  it. Then we can debate which one looks better. My guess is that we may  end up with a mix of the two.

I had a look and it's indeed the same app with 2 differences:
- in this PR, the Axle shim is used, the demo uses RxJava
- in this PR, the data access code is not internal to the entity.

As explained in the "Using Vert.x" guide, the Axle shim works nicely with the rest of Quarkus and MP API. So I believe it is better to use it in the guide.

As for data access code in the entity, if it's a blocker I'll move do the changes here, but I believe this style would better a fit a Reactive Panache guide.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
FroMage: Yes, I'm saying in quarkus for Hibernate ORM we put these statements in an `import.sql` file that is automagically executed at startup. We should probably do the same for the pg-client quarkus extension.

FroMage:
We should totally use Axle and I should update the http-rest-client demo to that too, agreed.

As for the SQL code location, I really think it should go in the entity to keep the model and controller logic separated.

using-vertx/src/main/java/org/acme/vertx/PgClientProducer.java
tsegismont: OK. Is it fine with you if `import.sql` contains both DML and DDD statements? In the Hibernate ORM extension it is different since they have a specific config property for drop and create schema.

using-vertx/src/main/resources/application.properties
```diff
@@ -1 +1,5 @@
...
+quarkus.datasource.url=vertx-reactive:postgresql:///quarkus_test
```
FroMage: Really three slashes?

using-vertx/src/main/java/org/acme/vertx/FruitResource.java
```diff
@@ -0,0 +1,100 @@
...
+                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Pear')"))
+                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Apple')"))
+                .toCompletableFuture()
+                .join();
+    }
+
+    @GET
+    public CompletionStage<Response> get() {
+        return Fruit.findAll(client)
+                .thenApply(fruits -> fruits.isEmpty() ? Response.noContent() : Response.ok(fruits))
```
FroMage: Same question about empty collections returned vs HTTP NO CONTENT as in the guide.

FroMage:
LGTM, just two minor questions.

tsegismont:
@FroMage I've updated the PR.

FroMage:
OK thanks. @cescoffier do you still have comments blocking this merge or shall we merge it?

using-vertx/src/main/java/org/acme/vertx/FruitResource.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.ws.rs.core.Response.ResponseBuilder;
+import javax.ws.rs.core.Response.Status;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.resteasy.annotations.jaxrs.PathParam;
+
+import io.reactiverse.axle.pgclient.PgPool;
+
+@Path("fruits")
+@Produces("application/json")
```
cescoffier: Should we use MediaType constants?

tsegismont: Done

cescoffier:
and it's merged!

tsegismont:
🍾🎉

And now the guide PR?

PR #131, opened 2019-04-02
Fix for hibernate-orm-resteasy QS

rsvoboda:
Fix for hibernate-orm-resteasy QS, was failing both in native and jvm mode


gsmet:
@rsvoboda Aren't they already removed in the `development` branch?

rsvoboda:
You are right, I didn't fetch latest changes.
But it is weird that the PR shows diff, PR is against quarkusio:development

PR #134, opened 2019-04-03
Cleanup in pom files - quarkus-arc, BOM versions

rsvoboda:
Cleanup in pom files - quarkus-arc, BOM versions

- removal of explicit quarkus-arc dependency definition
- removal of explicit version definitions to use BOM based version
- nimbus-jose-jwt version defined via property


gsmet:
Nice work, thanks!

Issue #136, opened 2019-04-08
Version Bump ignores getting-started/build.gradle


vorburger:
I just noticed that the 8af8b36ee17a9d396369b99505638029e9a55fa7 _[RELEASE] - Bump version to 0.13.0_ which @gsmet and @cescoffier did earlier today missed to bump the version from 0.12.0 to 0.13.0 in quarkus-quickstarts/getting-started/build.gradle 

I've raise PR #137 to manually fix it, but perhaps on the next bump whatever script is (presumably) used to do the release could be adjusted to do it automatically?

@stalep FYI

cescoffier:
This has been fixed.

PR #137, opened 2019-04-08
[RELEASE] bump version to 0.13.0 also for Gradle

vorburger:
see https://github.com/quarkusio/quarkus-quickstarts/issues/136

### DIFF
--- a/getting-started/build.gradle
+++ b/getting-started/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'java'
-    id 'io.quarkus.gradle.plugin' version '0.12.0'
+    id 'io.quarkus' version '0.13.0'
 }
 
 repositories {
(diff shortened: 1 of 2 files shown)


vorburger:
@gsmet @cescoffier @stalep

gsmet:
@cescoffier any chance you could take a look at automating it soon? I suspect we will have to release a 0.13.1 soon, considering the potential bootstrap issue we have.

vorburger:
@cescoffier I suspect you'll probably want to merge this (manual) PR already anyway 8because it doesn't just bump versions but also fixes IDs, which should be a one time thing), and THEN in #136 fix whatever script does the automated version bumps to do what I did manually here automatically (but simpler, forget about the IDs, just bump the version numbers).

cescoffier:
I don't know much Gradle, I fear it will end up with an error-prone sed / awk kind of thing (like https://stackoverflow.com/questions/42044160/gradle-equivalent-of-maven-versions-plugin).

cescoffier:
Looking at the gradle files right now, it must be refactored to avoid duplicating the versions:

```
plugins {
    id 'java'
    id 'io.quarkus.gradle.plugin' version '0.12.0' . <--- occurence 1
}

repositories {
     mavenCentral()
     maven { url 'http://repository.jboss.org/nexus/content/groups/public' }
}

dependencies {
    implementation group: 'io.quarkus', name: 'quarkus-resteasy', version:'0.12.0' <-- occurence 2

    testCompile group: 'io.quarkus', name: 'quarkus-junit5', version:'0.12.0' <-- occurence 3
    testCompile group: 'io.rest-assured', name: 'rest-assured', version: '3.3.0'
}
```

Can someone with gradle experience update the gradle files to use a single variable with a coherent name (so sed / awk magic can work).

gsmet:
Closing this one, we don't want to apply anything like that to master and have conflicts right away.

I'll extract the important bits of it and try to come with a scripted approach for the next release.

PR #139, opened 2019-04-08
add test { useJUnitPlatform } to getting-starting/build.gradle

vorburger:


vorburger:
This (currently) doesn't actually work, Gradle `test` fails, see https://github.com/quarkusio/quarkus/issues/1930 for full discussion, and is thus WIP and created for easy reproducibility by others, and to find how we want to address this best.

gsmet:
So we want that one but we can't merge it for now as it would break the build.

Please do not merge until quarkusio/quarkus#1930 is fixed!

vorburger:
With the 2nd new commit, we could already merge this PR as-is now, and then improve later, see https://github.com/quarkusio/quarkus/issues/1930.

gsmet:
I don't understand how we could merge that considering it's not working at the moment and the build fails?

vorburger:
No, as I said, with [the 2nd new commit](65dbabd34e3c306a723815034caacffbf0418f57) (which adds `exclude '**/*IT*'`), the build does not fail anymore with this PR as it is now. It will run least the non-native Gradle test, which is progress and a step in the right direction (currently it runs no tests at all).

vorburger:
Hello @gsmet I'm doing a year end clean up of my personal https://github.com/pulls and wanted to ask if you would still consider merging this very old Pull Request, or if we should forget about and close this? (This is a bulk message I'm posting to many issues to gauge what is still revelant.)

Issue #138, opened 2019-04-08
Add CI for PRs to QuickStarts


vorburger:
Perhaps it would be useful to validate all PRs proposed to this repo (Quick Starts) by CI?

May be that would caught https://github.com/quarkusio/quarkus/issues/1918 if https://github.com/quarkusio/quarkus-quickstarts/commit/8af8b36ee17a9d396369b99505638029e9a55fa7 was run through a PR with CI.

Such a CI job could also run the Gradle Getting Started (see #136).

Just a thought.

gsmet:
So, I'm all for getting back CI on quickstarts.

The issue is that we don't publish the snapshots right now so it's not as easy as it looks.

We run a full quickstart testing before releasing but the bootstrap issue apparently needs some special circumstances to happen (I still can't reproduce it).

For now, the version bump is directly pushed by a script so it's not dealt with as a PR but tests are run before pushing it.

So +1 but that needs work.

vorburger:
> The issue is that we don't publish the snapshots right now so it's not as easy as it looks.

@gsmet but we don't have to, or at least not just to enable CI, right? What I'm suggesting is really "just" CI for all PRs, nothing per se specific to version bump release processes, and without SNAPSHOT support...

Because the POMs in this repo (quickstarts) do NOT actually use 999-SNAPSHOT (may be they used to, but at least on current master they do not). So a future release version bump is "just another PR", which bumps from a (again, non-SNAPSHOT) say 0.13.0 to a 0.14.0 or whatever - which CI can verify - like any other change.

> We run a full quickstart testing before releasing 

That's probably also a Good Idea, but kind of separate and just a parallel pre-req to this suggestion? (I'm assuming that the script used to do this changes all versions to 999-SNAPSHOT; but that's temporary and not committed to this repo, or not anymore - it seems.)

cescoffier:
I've started working on it.

rsvoboda:
@cescoffier what's the status and overall plan ?

Will we publish snapshots or CI pipeline will build Quarkus latest bits ?
`mvn clean install -DskipTests -DskipITs` on Quarkus is quite fast

cescoffier:
I have started working on this, but we are facing a lot of connection issue with maven central and make our CI approach not really suitable for this.

cescoffier:
[...]

PR #140, opened 2019-04-09
Fix Gradle plugin id and parameterize the Quarkus version

gsmet:


vorburger:
FTR: This is inspired by #137, and related to #136 ... I'll provide some review feedback.

getting-started/build.gradle
```diff
@@ -1,16 +1,16 @@
 plugins {
     id 'java'
-    id 'io.quarkus.gradle.plugin' version '0.12.0'
+    id 'io.quarkus'
 }
 
 repositories {
+     mavenLocal()
```
vorburger: This really shouldn't be required? The build is more stable and reproducible without this, and we shouldn't promote needing it IMHO.

getting-started/gradle.properties
```diff
@@ -0,0 +1 @@
+quarkusVersion = 999-SNAPSHOT
```
vorburger: as per the discussion in #138, IMHO this is a bad idea, and should be `0.13.0` instead, just like in the Maven POM.

getting-started/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
vorburger: Having the Gradle wrapper duplicated inside each quickstart is not required IMHO. There already is one at the root, and [the documentation here](https://github.com/quarkusio/quarkus-quickstarts/blob/master/CONTRIBUTING.md#build-with-gradle) already suggests to use `../gradlew`. (I saw that the Maven wrapper is equally duplicated everywhere; if anything, I'd put a single Maven wrapper at the root instead as well.)

getting-started/build.gradle
gsmet: You can't build the `development` branch without it as we don't publish snapshots.

getting-started/gradle.properties
gsmet: It's for the development branch. It will be replaced automatically when releasing.

So for the next release, it will be automatically replaced by `0.14.0` when merged to master.

getting-started/gradle/wrapper/gradle-wrapper.properties
gsmet: Well, I had to look for it. So I'm pretty sure people will look for it too. That's why I added it.

getting-started/gradle.properties
vorburger: Oh, I've only just now figured that on the `development` branch it's `999-SNAPSHOT` even in the POMs, and the master branch has the released versions.. right, then it makes perfect sense.

getting-started/build.gradle
vorburger: so in an ideal world, this would be yanked by the script doing the releasing, but I guess that's minor

getting-started/gradle/wrapper/gradle-wrapper.properties
vorburger: Personally I hate seeing this many checked in JAR files, but anyway.

PR #154, opened 2019-04-15
feat(#153): Moves persistence quickstart from DMP to Testcontainers

### LINKED ISSUE #153: Persistence example moves from Docker Maven Plugin to Test Containers
With docker maven plugin you are tight to Maven lifecycle which makes painful to run tests from IDE. Ideally, these kinds of tests should come with an embedded lifecycle of persistence engine using TestContainer so everything is self-contained.

PR coming.

lordofthejars:
Fixes #153 byt adding TestContainers instead of depending on  Docker Maven Plugin which makes none direct run of tests from IDE.


hibernate-orm-panache-resteasy/src/test/java/FruitsEndpointTest.java
```diff
@@ -2,12 +2,31 @@
...
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
 
 @QuarkusTest
+@Testcontainers
 public class FruitsEndpointTest {
 
+    @Container
+    static PostgreSQLContainer DATABASE = new PostgreSQLContainer<>("postgres:10.5")
```
bsideup: Why not https://www.testcontainers.org/modules/databases/#database-containers-launched-via-jdbc-url-scheme ?:)

1. can be used as a static value somewhere in the test config
2. randomizes the ports
3. Much smaller change

lordofthejars: Although I agree that in a real project this would be interesting, as this is a quickstart I am not sure to add this on the JDBC Url. Basically, because of anyone who wants to start with Quarkus will also need to understand what is this `tc` thing part in URL. And should know that it is nothing related to Quarkus but for testing purposes. So since these quickstarters are used as tutorials, I'd advocate to not mix things.

lordofthejars: Saying that I like so much this feature.

bsideup: I see. Solid argument, agreed 👍

cescoffier:
@lordofthejars can you rebase?

lordofthejars:
@cescoffier Sorry don't ask me why but I get all these commits there when I synced with development.

cescoffier:
Weird set of commits, don't worry, I'm going t take care of this.

lordofthejars:
Maybe the best way is to just copy paste the code to a new PR and close this.

cescoffier:
I've migrated the changes to #213.  
Closing this one.

PR #156, opened 2019-04-15
Build minimal docker image by ldd and scratch.

philipz:
Using ldd to parse the java native execute file that needs those dependency of os library files, "copy_ldd.sh". Then building only execute file and os library files from Docker scratch image, and it just only 24MB.

### DIFF
--- a/getting-started-knative/Dockerfile
+++ b/getting-started-knative/Dockerfile
@@ -3,7 +3,7 @@ COPY . /project
 WORKDIR /project
 RUN mvn -Duser.home=/builder/home -B install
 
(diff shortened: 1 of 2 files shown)


getting-started-knative/Dockerfile
```diff
@@ -3,7 +3,7 @@ COPY . /project
 WORKDIR /project
 RUN mvn -Duser.home=/builder/home -B install
 
-FROM quay.io/quarkus/centos-quarkus-native-image:graalvm-1.0.0-rc15 as nativebuilder
+FROM swd847/centos-graal-native-image-rc12 as nativebuilder
```
Karm: Are you sure you wanted to change this?

philipz: Sorry! I missed it. I will fix it.

PR #157, opened 2019-04-15
Build minimal docker image by ldd and scratch.

philipz:
Using ldd to parse the java native execute file that needs those dependency of os library files, "copy_ldd.sh". Then building only execute file and os library files from Docker scratch image, and it just only 24MB.
images_size

### DIFF
--- a/getting-started-knative/Dockerfile
+++ b/getting-started-knative/Dockerfile
@@ -14,10 +14,12 @@ RUN  /opt/graalvm/bin/native-image -J-Djava.util.logging.manager=org.jboss.logma
      -H:-SpawnIsolates -H:-JNI --no-server -H:-UseServiceLoaderFeature -H:+StackTrace \
      && cp  -v quarkus-quickstart-knative-runner /tmp/quarkus-knative-runner
 
-FROM registry.fedoraproject.org/fedora-minimal
+FROM registry.fedoraproject.org/fedora-minimal as lddpackage
 RUN mkdir -p /work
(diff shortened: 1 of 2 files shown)


gsmet:
I think this would be an interesting blog post but I'm not sure having it in a quickstart makes sense as it's fairly advanced, possibly fragile, stuff.

philipz:
http://william-yeh.github.io/docker-mini/#1

rsvoboda:
I think small image is solved by https://github.com/quarkusio/quarkus-images/tree/graalvm-1.0.0-rc16/distroless

There are some details in https://github.com/quarkusio/quarkus/issues/2328#issuecomment-490086307 comment about the image size.

Agree with @gsmet that this is possibly fragile and quarkus-quickstarts are not the best place.

@philipz would you consider creating blogpost + tweeting about it ?

cescoffier:
A blog post about this would be awesome! We have a blog on the quarkus web site, we can host your post if you want.

(BTW the distroless build is also fragile, I did it ;-))

philipz:
Hi @cescoffier,
I putted a sample project, [https://github.com/philipz/quarkus-minimal](https://github.com/philipz/quarkus-minimal).
First, to run `./mvnw package -Pnative -Dquarkus.native.container-runtime=docker` to generate native image file.
Second, to run `docker build -f src/main/docker/Dockerfile philipz/quarkus-reactive .` to get the minimal docker image.
Maybe this tip could put  quarkus maven plugin or https://code.quarkus.io/.

PR #163, opened 2019-04-22
Separate Gradle getting started from the Maven one

AntoineLange:
Hello,

- Move getting-started-gradle to his own subdirectory

Fix [#2090](https://github.com/quarkusio/quarkus/issues/2090)


getting-started-gradle/src/test/java/quickstart/NativeGreetingResourceIT.java
```diff
@@ -0,0 +1,9 @@
+package quickstart;
+
+import io.quarkus.test.junit.SubstrateTest;
+
+@SubstrateTest
+public class NativeGreetingResourceIT extends GreetingResourceTest {
```
gsmet: Did you try to build the project with native? One of the reason we wanted a separate project is that, AFAIK, the IT tests should be in a separate directory and run separately from the unit tests.

gsmet: One of the reason it might not fail is because we haven't merged `https://github.com/quarkusio/quarkus-quickstarts/pull/139/files` yet.

See this comment here: https://github.com/quarkusio/quarkus/issues/1930#issuecomment-481179812 , especially about how to separate the tests.

We might also need to upgrade the Gradle project template in Quarkus itself.

gsmet:
@AntoineLange thanks for the changes. Don't we need:
```
test {
    useJUnitPlatform()
}
```
to actually get the tests running?

getting-started-gradle/src/test/java/unit/quickstart/GreetingResourceTest.java
```diff
@@ -0,0 +1,49 @@
...
+ *                    EXCLUSIVE AND CONFIDENTIAL PROPERTY OF
+ *
+ *                          EXPRETIO TECHNOLOGIES, INC.
+ *
+ * Any unauthorized use of this file or any of its parts, including, but not
+ * limited to, viewing, editing, copying, compiling, and distributing, is
+ * strictly prohibited.
+ *
+ * Copyright ExPretio Technologies, Inc., 2019. All rights reserved.
+ *=============================================================================
```
gsmet: Hmmm. This won't fly.

getting-started-gradle/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+    testCompile group: 'io.rest-assured', name: 'rest-assured', version: '3.3.0'
+}
+
+test {
+    useJUnitPlatform()
+}
+
+sourceSets {
+    test {
+        java.srcDir 'src/test/java/unit'
```
gsmet: I think I would have gone with `src/test/java` here.

getting-started-gradle/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+    }
+    integration {
+        java.srcDir 'src/test/java/integration'
```
gsmet: and `src/it/java` here.

getting-started-gradle/pom.xml
```diff
@@ -0,0 +1,119 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gsmet: Just above the out/ binary files are added. They shouldn't be. BTW, shouldn't they be in build/ as far as I understand the Gradle build?

getting-started-gradle/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+<project>
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.acme</groupId>
+    <artifactId>getting-started-gradle</artifactId>
+    <version>1.0-SNAPSHOT</version>
```
gsmet: We can drop the pom here. We want it to be a purely Gradle project. We can also remove the Maven wrapper from this directory if it's still there (haven't checked).

getting-started-gradle/src/main/java/quickstart/GreetingResource.java
```diff
@@ -0,0 +1,28 @@
+package quickstart;
```
gsmet: could we go with packages similar to the other projects? `org.acme.gradle` would be best. Same for the tests.

.gitignore
```diff
@@ -13,6 +13,7 @@ buildNumber.properties
 
 .gradle/
 build/
+out/
```
gsmet: Shouldn't everything be in `build/`? (I'm not so familiar with Gradle so it's a real question)

getting-started-gradle/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+    id 'java'
+    id 'io.quarkus'
+}
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+}
+
+dependencies {
```
gsmet: I think we could use our BOM, it would be better. Not sure how Gradle works with BOM, though?

getting-started-gradle/src/test/java/unit/quickstart/GreetingResourceTest.java
AntoineLange: Woops, sorry forget to switch context in IDE here.
I'm gonna work on this during the weekend.

getting-started-gradle/build.gradle
AntoineLange: Ok

getting-started-gradle/build.gradle
AntoineLange: Ok

getting-started-gradle/pom.xml
AntoineLange: Totally right

getting-started-gradle/pom.xml
AntoineLange: Sure

getting-started-gradle/src/main/java/quickstart/GreetingResource.java
AntoineLange: Ok :)

.gitignore
AntoineLange: Totally right. out/ was an IDE setting problem. I'm not so familiar with Gradle too tho.

emmanuelbernard:
Hey all, gentle nudge to see if we can hit home with this one :)

cescoffier:
@gsmet what's your opinion about this one?

PR #168, opened 2019-04-25
Fix #13 Introduce .editorconfig and editorconfig-maven-plugin

### LINKED ISSUE #13: Pom files should be indented consistently
Some of the pom files are indented with 2 spaces and some with 4 spaces. It makes it a bit tedious to copy/paste things from one to another.

I would vote for making it consistently 4 spaces and reformat everything.

It should be done when we don't have pull requests pending to avoid conflicts.

ppalaga:


cescoffier:
I was wondering if we should apply the same rules as the main repository?

ppalaga:
> I was wondering if we should apply the same rules as the main repository?

I can prepare a PR if there is enough hope for it to get accepted?

ppalaga:
Anyway, I'd appreciate any feedback on the present PR to know whether it is worth rebasing.

cescoffier:
I believe we should be consistent and having a single ruleset. It just makes our work easier (one IDE config). So, yes having a PR doing this would be highly appreciated.

cescoffier:
@gsmet Any opinion?

PR #184, opened 2019-05-03
Sync artifactId with directory name

rsvoboda:
Sync artifactId with directory name


rsvoboda:
Didn't find any reference in docs to changed QS artifactId(s)

PR #194, opened 2019-05-13
[using-keycloak] - policy-enforcer needs to be enabled explicitly

pedroigor:
See https://github.com/quarkusio/quarkus/issues/2422.

### DIFF
--- a/using-keycloak/src/main/resources/application.properties
+++ b/using-keycloak/src/main/resources/application.properties
@@ -5,6 +5,7 @@ quarkus.keycloak.auth-server-url=http://localhost:8180/auth
 quarkus.keycloak.resource=backend-service
 quarkus.keycloak.bearer-only=true
 quarkus.keycloak.credentials.secret=secret
+quarkus.keycloak.policy-enforcer.enable=true
 quarkus.keycloak.policy-enforcer.enforcement-mode=PERMISSIVE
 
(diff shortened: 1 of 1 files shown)


using-keycloak/src/main/resources/application.properties
```diff
@@ -5,6 +5,7 @@ quarkus.keycloak.auth-server-url=http://localhost:8180/auth
 quarkus.keycloak.resource=backend-service
 quarkus.keycloak.bearer-only=true
 quarkus.keycloak.credentials.secret=secret
+quarkus.keycloak.policy-enforcer.enable=true
```
gsmet: The title of your PR says it should be disabled but you enabled it?

BTW, is this something we want to disable by default in Quarkus itself?

Does the property just below make sense if you disable it?

pedroigor: The fix is based on https://github.com/quarkusio/quarkus/issues/2422. For this quickstart, in particular, we are enabling the `keycloak.policy-enforcer` configuration (which should be optional).

gsmet: Could you make the commit message clearer then? thanks!

pedroigor: @gsmet, I think the title now should be better ...

Issue #189, opened 2019-05-07
Gradle configs should enforce quarkus-bom versions


aloubyansky:
An example
```
dependencies {
    implementation enforcedPlatform("io.quarkus:quarkus-bom:${quarkusVersion}")
    implementation 'io.quarkus:quarkus-resteasy'
```
In this case the extension version will be coming from the BOM. And it will also use the correct versions for the transitive dependencies.
Currently, `org.wildfly.common:wildfly-common:1.5.0.Final` ends up being used instead of version `1.5.0.Final-format-001` which breaks some quickstarts.

PR #195, opened 2019-05-14
import Quarkus BOM to make sure the lib contains the right deps

aloubyansky:
Fix for #189


aloubyansky:
@gsmet in case you want to review it. I've verified it works for both the local snapshot and for released 0.14.0.

Issue #198, opened 2019-05-19
Native build fails for Camel quickstart


misl:
```
mvn clean verify -Pnative
```
Results in:

[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-failsafe-plugin must be a valid version but is '${surefire-plugin.version}'. @ line 95, column 22
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project org.acme:camel-java:1.0-SNAPSHOT (/home/misl/Dev/github/quarkus-quickstarts/camel-java/pom.xml) has 1 error
[ERROR]     'build.plugins.plugin.version' for org.apache.maven.plugins:maven-failsafe-plugin must be a valid version but is '${surefire-plugin.version}'. @ line 95, column 22
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException

PR #199, opened 2019-05-19
#198 Replaced surefire.version with surefire-plugin.version

misl:
- Replaced surefire.version with surefire-plugin version just like the other quickstarts

Fixed: #198


gsmet:
Thanks!

Issue #197, opened 2019-05-19
Create a quickstart using MQTT


misl:
There is quickstart for Kafka. A similar quickstart of MQTT will help people (like myself) get started with Quarkus and MQTT.

misl:
I have already started on taking the Kafka quickstart as an example for a MQTT quickstart. Pull request is coming.

jesuino:
Thanks for your great contribution, I saw your code in development! Do you know if this can be extended to use AWS IoT? Would you suggest some example using AWS IoT?

misl:
Hi jesuino,

Unfortunately I am not familiar with AWS IoT, but probably there are similarities with Google IoT and IBM Watson IoT. Even though it is MQTT, each has there own topic naming convention you have to follow (as a result of there device management). This makes it hard to migrate (vendor lock-in).

Getting back to your question. If AWS IoT is based on MQTT it should work. 
- Just take the quickstart example
- configure it to connect with AWS IoT 
  - set the right credentials 
  - subscribe and public to the right topics (Check topic naming convention with AWS IoT)

and give it a go.

misl:
This issue can closed since the changes were merged

jesuino:
Thanks, @misl was just wondering if you tested it with AWS IoT. Unfortunately yes, I agree with you, they provide a Java API which help us to set certificates on their client...

misl:
So what you are actually looking for is support for certificates? Have a look at [smallrye](https://smallrye.io/) (the library taking care of micro profile messaging for Quarkus). Is seems smallrye's [MqttSource](https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java) does not yet support certificates. However, the underlying Vert.x (have a look at [MqttClientOptions](https://vertx.io/docs/apidocs/io/vertx/mqtt/MqttClientOptions.html)) does seems to have support for certificates. 

So it seems adding support for certificates might not be too complicated. I suggest you create an issue and request support for certificates. I am quite sure they accept pull requests (they already accepted some of mine).

PS. maybe this conversation should be part of a Smallrye messaging issue/ticket.

PR #200, opened 2019-05-19
#197 Added quickstart for MicroPofile messaging using MQTT

misl:
- Add new module microprofile-messaging-mqtt
- The new module was also add to the README

Fixed: #197 

Notes:
- Lacks unittests. Just like Kafka quickstart which begin used as the reference implementation .
- Lacks integration tests. Again just like Kafka quickstart.
- Personally I don't like the smallrye-reactive-messaging-mqtt version 0.0.6 dependency, probably need a proper extension to get rid of this.

PS: I lack the Quarkus experience to know the best way to add any of these. Any help would be greatly appreciated.


microprofile-messaging-mqtt/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+      <artifactId>quarkus-smallrye-reactive-messaging</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-vertx</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-reactive-messaging-mqtt</artifactId>
+      <version>0.0.6</version>
```
cescoffier: Not related to this PR, but I wanted to create a specific extension for MQTT.

misl: I agree on creating an extension I already gave it a small try myself, but I am not under the impression that I know what I am doing. I am not yet familiar with the Quarkus extension mechanism.

PR #206, opened 2019-05-28
Hibernate search elasticsearch

gsmet:
Quickstart associated with the guide contributed here: https://github.com/quarkusio/quarkus/pull/2624 .


gsmet:
@yrodiere interested in your feedback here too.

hibernate-search-elasticsearch/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,284 @@
...
+    </div>
+    <div class="ui modal" id="modal-author">
+        <i class="close icon"></i>
+        <div class="header">
+            {{ authorModalTitle }}
+        </div>
+        <div class="content">
+            <div class="description">
+                <input type="hidden" name="authorId" ng-model="form.author.id" />
+                <form class="ui form">
```
yrodiere: For some reason, pressing "enter" here in the input boxes doesn't submit the form. Not a big deal, but I though you might want to know.

hibernate-search-elasticsearch/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,284 @@
...
+        <div class="header">
+            Add book
+        </div>
+        <div class="content">
+            <div class="description">
+                <input type="hidden" name="authorId" ng-model="form.book.authorId" />
+                <form class="ui form">
+                    <div class="required field">
+                        <label>Title</label>
+                        <input type="text" name="title" placeholder="Title" ng-model="form.book.title" required>
```
yrodiere: Same here, pressing "enter" here doesn't submit the form.

hibernate-search-elasticsearch/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,284 @@
...
+                        <label>Last name</label>
+                        <input type="text" name="lastName" placeholder="Last name" ng-model="form.author.lastName" required>
+                    </div>
+                </form>
+            </div>
+        </div>
+        <div class="actions">
+            <button class="ui positive right labeled icon button" ng-click="addOrEditAuthor()">
+                <i class="checkmark icon"></i>
+                Submit
```
yrodiere: When adding an author and the list is filtered by a search pattern, the author might not show up. Maybe you could add a notifications that confirms the author was created and reminds it might not appear because of search criteria?

hibernate-search-elasticsearch/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,284 @@
...
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+    <div class="ui container search">
+        <form class="ui form" ng-submit="refresh()">
+            <div class="fluid big left ui action input">
+                <input type="text" placeholder="Search for an author name or a book title. Try searching for aus*, owen or mystery." ng-model="pattern">
+                <button class="ui blue right labeled icon button" ng-click="refresh()"><i class="search icon"></i> Search</button>
```
yrodiere: It would be great to also grey out books that do not match the search pattern. It would only require a second search query, and I think would make the demo even more appealing.

yrodiere: Also, I think removing the button and updating the list after a slight delay without typing (~200ms) would feel better. It's not like we have to care about overloading the server :) Not sure how easy that would be, though.

hibernate-search-elasticsearch/.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,117 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import java.net.*;
+import java.io.*;
+import java.nio.channels.*;
+import java.util.Properties;
+
+public class MavenWrapperDownloader {
```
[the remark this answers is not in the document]
gsmet: It's supposed to be part of the PR. We want the Maven wrapper to be available in the project directory.

hibernate-search-elasticsearch/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+This is a CRUD service exposing a couple of endpoints over REST,
+with a front-end based on Angular so you can play with it from your browser.
+
+While the code is surprisingly simple, under the hood this is using:
+ - RESTEasy to expose the REST endpoints
+ - Hibernate ORM to perform the CRUD operations on the database
+ - Hibernate Search + Elasticsearch to index the entities in an Elasticsearch index
+ - A PostgreSQL database; see below to run one via Docker
+ - ArC, the CDI inspired dependency injection tool with zero overhead
```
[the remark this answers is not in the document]
gsmet: This part is coming from Sanne's Hibernate ORM demo, I haven't changed it. Let's keep it this way. We can still revisit all of them later.

hibernate-search-elasticsearch/README.md
```diff
@@ -0,0 +1,123 @@
...
+You should then use this JDK to run the Maven build.
+
+## Building the demo
+
+After having set GraalVM as your JVM, launch the Maven build on
+the checked out sources of this demo:
+
+> mvn package
+
+## Running the demo
```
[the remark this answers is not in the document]
gsmet: Yeah, I don't know. The whole point of these quickstarts is to be companions for the guides. So we want people to understand what they are doing as that's how we learn.

Maybe we need a shortcut for people that just want a quick look but, yes, that's a global discussion as we have the issue with most quickstarts.

hibernate-search-elasticsearch/src/main/java/org/acme/hibernate/search/elasticsearch/LibraryResource.java
```diff
@@ -0,0 +1,124 @@
...
+    public void addBook(@FormParam("title") String title, @FormParam("authorId") Long authorId) {
+        Author author = Author.findById(authorId);
+        if (author == null) {
+            return;
+        }
+
+        Book book = new Book();
+        book.title = title;
+        book.author = author;
+        book.persist();
```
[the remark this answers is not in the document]
gsmet: Yes, bytecode enhancement is enabled and the code here should be updated to use the special Hibernate methods. So I think we are safe.

hibernate-search-elasticsearch/src/main/java/org/acme/hibernate/search/elasticsearch/model/Author.java
```diff
@@ -0,0 +1,53 @@
...
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.IndexedEmbedded;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.KeywordField;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+
+@Entity
+@Indexed
+public class Author extends PanacheEntity {
+
+    @FullTextField(analyzer = "defaultAnalyzer")
```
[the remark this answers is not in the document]
gsmet: I see your point. But, what I think will happen in real life: people will have a default one and several other specific ones for specific purposes.

That being said, I can certainly improve the names, I'll work on it.

hibernate-search-elasticsearch/src/main/resources/application.properties
```diff
@@ -0,0 +1,30 @@
...
+
+quarkus.datasource.url=jdbc:postgresql:quarkus_test
+quarkus.datasource.driver=org.postgresql.Driver
+quarkus.datasource.username=quarkus_test
+quarkus.datasource.password=quarkus_test
+
+quarkus.hibernate-orm.database.generation=drop-and-create
+
+quarkus.hibernate-search.elasticsearch.version=7
+quarkus.hibernate-search.elasticsearch.analysis-configurer=org.acme.hibernate.search.elasticsearch.config.AnalysisConfigurer
```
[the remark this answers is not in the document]
gsmet: It does not at the moment. That's on my TODO list.

hibernate-search-elasticsearch/src/main/resources/META-INF/resources/index.html
gsmet: Yeah, so we need to be very careful about what we do. The demo as it is is already on the complex side when trying to follow the guide. Keep in mind that's this demo is really a companion for the guide.

We could improve things but it will come with the cost of added complexity for the guide, which wouldn't be a good idea IMHO.

hibernate-search-elasticsearch/src/main/resources/META-INF/resources/index.html
gsmet: I was wondering if I would be the only one bothered by this one :). I'll fix it.

hibernate-search-elasticsearch/src/main/resources/META-INF/resources/index.html
gsmet: I now reset the search.

hibernate-search-elasticsearch/src/main/java/org/acme/hibernate/search/elasticsearch/model/Book.java
```diff
@@ -17,8 +17,8 @@
 @Indexed
 public class Book extends PanacheEntity {
 
-    @FullTextField(analyzer = "stemmingAnalyzer")
-    @KeywordField(name = "title_sort", sortable = Sortable.YES, normalizer = "defaultNormalizer")
+    @FullTextField(analyzer = "english")
+    @KeywordField(name = "title_sort", sortable = Sortable.YES, normalizer = "sort")
```
yrodiere: I think you removed "title_sort" from the guide?

hibernate-search-elasticsearch/README.md
```diff
@@ -80,7 +83,10 @@ When you're done playing with "dev-mode" you can run it as a standard Java appli
 
 First compile it:
 
-> mvn package
+> ./mvnw package
+
+Note that this command will start a PostgreSQL instance and an Elasticsearch cluster to execute the tests.
```
yrodiere: I don't know if you changed something, but from what I understand:

* PostgreSQL and Elasticsearch are started in the "test-compile" and "process-test-classes" phases
* then LibraryResourceTest, which is an integration test masquerading as a unit test, is run by surefire if not in native mode
* then LibraryResourceInGraalIT, which is an integration test, is run by failsafe if in native mode

I find this very strange and I would expect the following:

* PostgreSQL and Elasticsearch are started in the "pre-integration-test" phase
* then LibraryResourceTest, rename to LibraryResourceIT, is run by failsafe if not in native mode
* then LibraryResourceInGraalIT, which is an integration test, is run by failsafe if in native mode

If you did that, then `./mvnw package` would not run the integration tests anymore. You could use `./mvnw verify` if you need that.

hibernate-search-elasticsearch/src/main/java/org/acme/hibernate/search/elasticsearch/model/Book.java
gsmet: Ah yes, good catch.

hibernate-search-elasticsearch/README.md
gsmet: That's how it's done everywhere else so I won't change that.

PR #214, opened 2019-06-06
ubi8 based container for native mode application

rsvoboda:
ubi8 based container for native mode application

Followup of https://github.com/quarkusio/quarkus/pull/2723

Used command: `git grep -l fedora-minimal | xargs sed -i '' "s,registry.fedoraproject.org/fedora-minimal,registry.access.redhat.com/ubi8/ubi-minimal,g"`

### DIFF
--- a/application-configuration/src/main/docker/Dockerfile.native
+++ b/application-configuration/src/main/docker/Dockerfile.native
@@ -14,7 +14,7 @@
 # docker run -i --rm -p 8080:8080 quarkus/application-configuration
 #
 ###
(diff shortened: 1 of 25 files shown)


gsmet:
So did we decide to go with ubi8 image, even considering the libcrypt issue?

rsvoboda:
libcrypt issue is Fedora 30 only issue

If Fedora 30 people build using `-Dnative-image.docker-build=true` there is no libcrypt issue

PR #226, opened 2019-06-19
#2663 Misc. clean-up

gunnarmorling:
* Consistent default Kafka host
* Removing superfluous dependency
* Metadata exposed as JSON

@cescoffier Some more clean-up for the KStreams quickstart. Once https://github.com/quarkusio/quarkus/pull/2880 is merged, I'll send another one for removing the SerDe from the quickstart.


gsmet:
It really starts to be late to do more adjustments. I have to run the full quickstart build with native before 2pm and it takes 2 hours to run...

gsmet:
@gunnarmorling I merged quarkusio/quarkus#2880, can you make the adjustments you were talking about?

gunnarmorling:
Pushed one more commit for getting rid of the local Serde and using the one from the Kafka client module 🎉

gsmet:
I haven't run the full quickstart but I checked everything compiles, native included.

@gunnarmorling I will merge this one but we will need to have these quickstarts included in the build. Right now, they are not as the modules are not declared in the parent pom.

gsmet:
(I didn't do it myself as I thought you might have had a good reason to not do it)

gunnarmorling:
No good reason apart from not knowing about that aggregator POM :) So if
you have the bandwidth, feel free to add this (a parent for the two
probably would make sense).

PR #231, opened 2019-06-24
PgRowSet is now an Iterable<Row> in Axle shim

tsegismont:
So using the PgIterator is no longer needed in the quickstart.


tsegismont:
@FroMage @cescoffier can you please review?

cescoffier:
@tsegismont can you check there is no change required in the doc itself?

PR #234, opened 2019-06-30
Extend Rest client quickstart to mention the CompletionStage support

machi1990:
See https://github.com/quarkusio/quarkus/issues/2119#issuecomment-484781468

Follows up https://github.com/quarkusio/quarkus/pull/3026 which adds rest client async calls test and docs docs to the guide.

### DIFF
--- a/rest-client/src/test/java/org/acme/restclient/CountriesResourceTest.java
+++ b/rest-client/src/test/java/org/acme/restclient/CountriesResourceTest.java
@@ -14,7 +14,7 @@ public class CountriesResourceTest {
         given()
           .when().get("/country/name/greece")
           .then()
-             .statusCode(200)  
+             .statusCode(200)
              .body("$.size()", is(1),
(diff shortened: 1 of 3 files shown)


rest-client/src/test/java/org/acme/restclient/CountriesResourceTest.java
```diff
@@ -23,4 +23,17 @@ public void testCountryNameEndpoint() {
              );
     }
 
-}
\ No newline at end of file
+    @Test
+    public void testCountryNameAsyncEndpoint() {
+        given()
+          .when().get("/country/name/greece")
```
cescoffier: You are not calling the right endpoint here? Shouldn't it be name-async?

machi1990: Good catch. It has been updated.

cescoffier:
Thanks!

PR #238, opened 2019-07-07
Update MicroProfile Health quickstart to MicroProfile Health 2.0

xstefank:


microprofile-health/pom.xml
```diff
@@ -5,15 +5,14 @@
   <groupId>org.acme</groupId>
   <artifactId>microprofile-health</artifactId>
   <version>1.0-SNAPSHOT</version>
-
```
gsmet: Let's keep that one.

microprofile-health/pom.xml
```diff
@@ -5,15 +5,14 @@
   <groupId>org.acme</groupId>
   <artifactId>microprofile-health</artifactId>
   <version>1.0-SNAPSHOT</version>
-  
   <properties>
-    <quarkus.version>999-SNAPSHOT</quarkus.version>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <surefire-plugin.version>2.22.1</surefire-plugin.version>
+    <quarkus.version>999-SNAPSHOT</quarkus.version>
```
gsmet: Let's keep the quarkus.version first and have the 2 encodings together.

microprofile-health/pom.xml
```diff
@@ -45,8 +39,11 @@
       <artifactId>rest-assured</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-health</artifactId>
```
gsmet: Let's put that one at the top.

(I suppose you have generated the file again but it's more readable for the user this way so I think it's worth handcrafting a few things)

PR #246, opened 2019-07-21
Upgrade Gradle version to 5.5.1 (includes enforcedPlatform bug fix)

gwenneg:
Version `5.5` fixed an `enforcedPlatform` bug: https://github.com/gradle/gradle/issues/9424

The `gradle-wrapper.jar` file's checksum can be verified using [this procedure](https://docs.gradle.org/current/userguide/gradle_wrapper.html#configuring_checksum_verification) and [the checksum value from Gradle's website](https://gradle.org/release-checksums/).


gsmet:
Merged, thanks!

PR #248, opened 2019-07-25
Update FruitResource.java

blue61:
added GET commands to actually Create, Update or Delete Entities, as the other HTTP Verbs are often unaccessible in browsers.
GET /fruits/-/name    creates a new entity
GET /fruits/id/name   updates the entity with the given id to have the given name
GET /fruits/id/-          deletes the entity with the given id

### DIFF
--- a/hibernate-orm-panache-resteasy/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
+++ b/hibernate-orm-panache-resteasy/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
@@ -42,6 +42,39 @@ public class FruitResource {
         return entity;
     }
 
+    @GET
+    @Path("-/{name}")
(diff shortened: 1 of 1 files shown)


cescoffier:
Not sure about that one.

@FroMage WDYT?

FroMage:
> the other HTTP Verbs are often unaccessible in browsers

Are you sure? XmlHttpRequest supports it for ever, I think.

`GET` requests should be idempotent and should be cacheable and never ever modify data, so I think this is really not the right way to do REST.

gsmet:
I haven’t merged that one because I thought it’s better to have the demo as simple as possible and not add too many things. 

It’s just my preference though.

cescoffier:
I believe we should not merge. It does not follow the REST good practices,
and it complexifies the QuickStart.

PR #249, opened 2019-07-30
Add test for kafka-quickstart

### LINKED ISSUE #145: kafka-quickstart has no test coverage
kafka-quickstart has no test coverage

kmos:
Hi, 

this PR add test for kafka-quickstart.

Resolves #145 

- avoid to use kafka in test profile
- add resteasy-client

### DIFF
--- a/kafka-quickstart/src/main/resources/application.properties
+++ b/kafka-quickstart/src/main/resources/application.properties
@@ -1,9 +1,21 @@
 # Configure the Kafka sink (we write to it)
-mp.messaging.outgoing.generated-price.connector=smallrye-kafka
-mp.messaging.outgoing.generated-price.topic=prices
-mp.messaging.outgoing.generated-price.value.serializer=org.apache.kafka.common.serialization.IntegerSerializer
+%dev.mp.messaging.outgoing.generated-price.connector=smallrye-kafka
+%dev.mp.messaging.outgoing.generated-price.topic=prices
+%dev.mp.messaging.outgoing.generated-price.value.serializer=org.apache.kafka.common.serialization.IntegerSerializer
 
 # Configure the Kafka source (we read from it)
-mp.messaging.incoming.prices.connector=smallrye-kafka
-mp.messaging.incoming.prices.topic=prices
-mp.messaging.incoming.prices.value.deserializer=org.apache.kafka.common.serialization.IntegerDeserializer
+%dev.mp.messaging.incoming.prices.connector=smallrye-kafka
+%dev.mp.messaging.incoming.prices.topic=prices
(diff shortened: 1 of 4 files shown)


kafka-quickstart/src/test/java/org/acme/quarkus/sample/PriceResourceTest.java
```diff
@@ -0,0 +1,39 @@
...
+    void testPricesEventStream() {
+        Client client = ClientBuilder.newClient();
+        WebTarget target = client.target(PRICES_SSE_ENDPOINT);
+
+        List<Double> received = new ArrayList<>();
+
+        try (SseEventSource source = SseEventSource.target(target).build()) {
+            source.register(inboundSseEvent -> received.add(Double.valueOf(inboundSseEvent.readData())));
+            source.open();
+            Thread.sleep(500); // Consume events for just 500 ms
```
cescoffier: Can you switch to Awaitility instead? 

```
await()atMost(10000).until(() -> received.size() < 5)
```

`received` would need to be a `CopyOnWritearrayList` but that would be more robust.

kafka-quickstart/src/main/resources/application.properties
```diff
@@ -1,9 +1,21 @@
 # Configure the Kafka sink (we write to it)
-mp.messaging.outgoing.generated-price.connector=smallrye-kafka
-mp.messaging.outgoing.generated-price.topic=prices
-mp.messaging.outgoing.generated-price.value.serializer=org.apache.kafka.common.serialization.IntegerSerializer
+%dev.mp.messaging.outgoing.generated-price.connector=smallrye-kafka
```
cescoffier: Any reason for this change?
I would understand if dev and prod configs would be different, but it's not the case here.

cescoffier: Oh I see, in `test` you generate only 3 values.

kmos: Yes. For what I understand here quarkusio/quarkus/issues/2548 the main idea is to detect problems in dependencies. There is no need to have a real istance of kafka. 

I was obblied to duplicate the values in order to avoid configuration of kafka sink for ```test``` profile (but leave it for dev and prod)

kafka-quickstart/src/test/java/org/acme/quarkus/sample/PriceResourceTest.java
kmos: addressed, great suggestion!

kmos:
What can I do to solve this PR? I have addressed the change requested. @cescoffier

PR #260, opened 2019-08-22
Make Postgresql host explicit

galderz:
Aside from this, I was wondering if it'd be useful to explain in the corresponding READMEs how these ORM examples could be run within a Kubernetes environment?

I was trying to run both the quickstart and DB in Minikube, and that's what lead me down this path. In such env the host needs to be aligned to the DB service name, hence the importance of making this explicit.

I ended up creating a Kubernetes specific [Dockerfile](https://github.com/galderz/putcachejpa/blob/master/src/main/docker/Dockerfile.kubernetes#L22) that passed in database URL adjusted to the DB Kubernetes service. Use [this Makefile](https://github.com/galderz/putcachejpa/blob/master/Makefile) to drive building and running app and DB in kubernetess. 

If added, could this be easily tested too?

### DIFF
--- a/hibernate-orm-panache-resteasy/README.md
+++ b/hibernate-orm-panache-resteasy/README.md
@@ -109,3 +109,14 @@ Navigate to:
 <http://localhost:8080/index.html>
 
 Have fun, and join the team of contributors!
+
+## Running the demo in Kubernetes
+
+This section provides extra information for running both the database and the demo on Kubernetes.
+As well as running the DB on Kubernetes, a service needs to be exposed for the demo to connect to the DB.
+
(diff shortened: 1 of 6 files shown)


cescoffier:
@galderz did you change the guide? (adding an admonition might make sense no?)

galderz:
@cescoffier Not yet, but I'll add a note now.

galderz:
@cescoffier Added a note in each of the quickstarts.

cescoffier:
How missed on big details. You must target the `development` branch, not the `master` branch. I believe you need to re-open the PR.

gsmet:
@galderz can you please provide a PR against the `development` branch? Thanks!

PR #261, opened 2019-08-23
Add DynamoDB quickstart

marcinczeczko:
Added DynamoDB quickstart related to the https://github.com/quarkusio/quarkus/pull/3172
At the moment it uses quarkus SNAPSHOT version. To be updated and ready for review/merge once the quarkus extension to be released.


amqp-quickstart/pom.xml
```diff
@@ -8,7 +8,7 @@
   <properties>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <surefire-plugin.version>2.22.0</surefire-plugin.version>
-    <quarkus.version>999-SNAPSHOT</quarkus.version>
```
cescoffier: Qucisktarts on PR must target the development branch and use 999-SNAPSHOT. The bump to the release version happen during the release process.

dynamodb-quickstart/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-bom</artifactId>
+                <version>${quarkus.version}</version>
+                <type>pom</type>
+
```
cescoffier: Extra blank line can be removed

dynamodb-quickstart/src/main/java/org/acme/dynamodb/FruitAsyncService.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public CompletableFuture<List<Fruit>> add(Fruit fruit) {
+        return dynamoDB.putItem(putRequest(fruit)).thenCompose(ret -> findAll());
+    }
+
+    public CompletableFuture<Fruit> get(String name) {
+        return dynamoDB.getItem(getRequest(name)).thenApply(resp -> Fruit.from(resp.item()));
+    }
+
+    //    private static CompletableFuture<Boolean> retryAsync(Supplier<CompletableFuture<DescribeTableResponse>> action,
```
cescoffier: Why is this commented out?

dynamodb-quickstart/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,152 @@
+<!DOCTYPE html>
```
cescoffier: Don't include the default landing page in quickstarts

dynamodb-quickstart/src/main/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+quarkus.dynamodb.region=eu-central-1
```
cescoffier: I'm a dynamodb noob, but I'm wondering why you need to specify a region if the endpoint is targeting localhost?

dynamodb-quickstart/src/main/java/org/acme/dynamodb/FruitAsyncService.java
marcinczeczko: Removed - it was left over

dynamodb-quickstart/src/main/resources/META-INF/resources/index.html
marcinczeczko: removed

dynamodb-quickstart/src/main/resources/application.properties
marcinczeczko: If region is not specified, then default region provider chain fires that looks up for region in env properites, system properties or aws profile files on local file system. Basically even local instance require region to be specified as it's part of the HTTP request to the dynamodb (local instance requires it too, but it can be any region)

amqp-quickstart/pom.xml
marcinczeczko: Thanks. Just realised it when looking at the repo.

dynamodb-client/src/main/java/org/acme/dynamodb/FruitAsyncResource.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/async-fruits")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class FruitAsyncResource {
+
+    @Inject
+    FruitAsyncService service;
```
gsmet: If I'm correct, the async support was disabled for now? The quickstart should take that into account.

It should probably be stated in the guide and code should be commented here so that it works. I'm not asking to remove the code, just adjust so that the build doesn't fail.

marcinczeczko: I was about to do so, but yesterday evening AWS SDK released a version with Netty 4.1.39. See https://github.com/quarkusio/quarkus/pull/3839 that re-enables async support.

PR #263, opened 2019-08-27
Pipelines - increase timeoutInMinutes

rsvoboda:
Increase `timeoutInMinutes` for pipelines, build are cancelled after 1 hour.

Number of quickstarts increased a bit and new are coming.


rsvoboda:
@stuartwdouglas can we have CI runs for PRs ?

stuartwdouglas:
Sure

PR #264, opened 2019-08-29
Fix AnalysisConfigurer to implement updated ElasticsearchAnalysisConfigurer

rsvoboda:
Fix AnalysisConfigurer to implement updated ElasticsearchAnalysisConfigurer.

Quickstart can't be compiled without this change ...


gsmet:
I was going to review the quickstarts for the release, thanks for doing it :).

rsvoboda:
I saw same fails in NATIVE few days ago, wanted to check again

PR #269, opened 2019-08-29
Add Spring Data JPA guide

geoand:


using-spring-data-jpa/src/main/java/org/acme/spring/data/jpa/FruitResource.java
```diff
@@ -0,0 +1,62 @@
...
+            return fruitRepository.save(fruit);
+        }
+
+        throw new IllegalArgumentException("No Fruit with id " + id + " exists");
+    }
+
+    @GET
+    @Path("/color/{color}")
+    @Produces("application/json")
+    public List<Fruit> all(@PathParam("color") String color) {
```
gsmet: I don't remember what was the outcome of our discussion about the name of this method?

using-spring-data-jpa/src/main/resources/application.properties
```diff
@@ -0,0 +1,7 @@
+quarkus.datasource.url=jdbc:postgresql:quarkus_test
```
gsmet: Apparently, people would prefer we specify the host: https://github.com/quarkusio/quarkus-quickstarts/pull/260/files#diff-284b129d02c2070784925ee9ac16a157R1

using-spring-data-jpa/src/main/java/org/acme/spring/data/jpa/FruitResource.java
geoand: You are right, as usual :) 
We had changed it, I'll fix it

using-spring-data-jpa/src/main/resources/application.properties
geoand: Interesting, I'll update then :)

geoand:
@gsmet Thanks for checking!

PR updated as per your suggestions

PR #272, opened 2019-08-31
Add Rest Json with static resource

lbernardomaia:
The idea of this quickstart is to show how to use a static file using a Native Image, as an additional configuration is required to work.


gsmet:
Interesting. @cescoffier WDYT about this one? I'm not sure it's worth a separate quickstart but I'm not sure either complicating an existing one is a good idea.

We have some documentation about this here: https://quarkus.io/guides/writing-native-applications-tips

cescoffier:
It would worth a full quickstart when we would have the _cache_ config support. 
We can add it as a new one and extend it later.

PR #277, opened 2019-09-11
Fix race condition in reactive add

gcollin:
Where sometimes the list is run before the add is taking place.


gcollin:
Sorry wrong request

PR #278, opened 2019-09-11
Fix race condition in reactive add

gcollin:
Where sometimes the list is run before the add is taking place.

### DIFF
--- a/using-mongodb-client/src/main/java/org/acme/rest/json/ReactiveFruitResource.java
+++ b/using-mongodb-client/src/main/java/org/acme/rest/json/ReactiveFruitResource.java
@@ -22,6 +22,6 @@ public class ReactiveFruitResource {
 
     @POST
     public CompletionStage<List<Fruit>>  add(Fruit fruit) {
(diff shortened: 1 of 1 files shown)


gsmet:
@cescoffier can you have a look at this one?

using-mongodb-client/src/main/java/org/acme/rest/json/ReactiveFruitResource.java
```diff
@@ -22,7 +22,6 @@
 
     @POST
     public CompletionStage<List<Fruit>>  add(Fruit fruit) {
-        fruitService.add(fruit);
-        return list();
+        return fruitService.add(fruit).thenCompose(aVoid -> list());
```
cescoffier: I would use `x` instead of `aVoid`.

gcollin: Sure.

gcollin: done

PR #283, opened 2019-09-17
Move using-spring-web tests to the same package as sources

rsvoboda:
Move using-spring-web tests to the same package as sources


gsmet:
Thanks!

PR #286, opened 2019-09-18
fix: initial version of react frontend

edewit:
fixing https://github.com/quarkusio/quarkus/issues/588

Map JAX-RS under a specific path (/api), then configure the frontend server to proxy to quarkus. When doing a maven build the frontend gets build by the maven-frontend-plugin and added to quarkus.


edewit:
@gsmet is this something worth including?

gastaldi:
I think this is pretty valid and a common request among Quarkus users. WDYT @gsmet @cescoffier @emmanuelbernard ?

PR #291, opened 2019-09-24
Add Artemis JMS example

middagj:
See quarkusio/quarkus#4186 for a pull request for the Artemis guide in Quarkus.

If Artemis is upgraded in [Quarkus](quarkusio/quarkus#4184), the artemis-server probably needs to be upgraded to 2.10.0.


artemis-jms/src/main/java/org/acme/artemis/PriceConsumer.java
```diff
@@ -0,0 +1,57 @@
...
+        scheduler.shutdown();
+    }
+
+    @Override
+    public void run() {
+        try (JMSContext context = connectionFactory.createContext(Session.AUTO_ACKNOWLEDGE)) {
+            JMSConsumer consumer = context.createConsumer(context.createQueue("prices"));
+            while (true) {
+                Message message = consumer.receive();
+                if (message == null) return;
```
cescoffier: Can you add a comment explaining that `message==null` is a termination signal?

artemis-jms/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,152 @@
+<!DOCTYPE html>
```
cescoffier: Can you drop this page (while generated, we don't want it in the quickstart code).

artemis-jms/src/test/resources/broker.xml
```diff
@@ -0,0 +1,23 @@
+<configuration xmlns="urn:activemq" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:activemq /schema/artemis-configuration.xsd">
+    <core xmlns="urn:activemq:core">
+        <paging-directory>./target/artemis/paging</paging-directory>
+        <bindings-directory>./target/artemis/bindings</bindings-directory>
+        <journal-directory>./target/artemis/journal</journal-directory>
+        <large-messages-directory>./target/artemis/large-messages</large-messages-directory>
+
+        <connectors>
```
cescoffier: More a proposal for the extension itself.
Would it be possible to configure the extension from the `application.properties` file?

middagj: This is the server configuration. The extension can only configure the client aspects. As far as I know, large part of the configuration is only configurable in the server.

artemis-jms/src/main/resources/META-INF/resources/index.html
middagj: Sure, should I also drop the rest?

artemis-jms/src/main/java/org/acme/artemis/PriceConsumer.java
middagj: Yes, good idea.

artemis-jms/src/test/resources/broker.xml
cescoffier: Oh sorry, my bad!

PR #293, opened 2019-09-25
MongoDB with Panache quickstart

loicmathieu:
The MongoDB with Panache quickstart application.


gsmet:
Could you rename it to simply `mongodb-panache`? I think I would like to get rid of the `using-` prefix at some point.

using-mongodb-panache/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+  </dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy-jsonb</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-mongodb-panache</artifactId>
+      <version>999-SNAPSHOT</version>
```
gsmet: That shouldn't be here. If it's generated with this then we have an issue with the bom.

using-mongodb-panache/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,152 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>using-mongodb-panache - 1.0-SNAPSHOT</title>
```
gsmet: We would drop the `using-` from here too.

using-mongodb-panache/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+            <executions>
+              <execution>
+                <goals>
+                  <goal>integration-test</goal>
+                  <goal>verify</goal>
+                </goals>
+                <configuration>
+                  <systemProperties>
+                    <native.image.path>${project.build.directory}/${project.build.finalName}-runner</native.image.path>
+                  </systemProperties>
```
gsmet: Will the tests run properly? I don't see anything to launch a MongoDB instance for the tests.

using-mongodb-panache/pom.xml
loicmathieu: It's a chicken an egg issue, the quickstarts are written before the extension goes into Quarkus so we need to hardcode the version, almost all guides does this.
If I remove it, you will not be able to merge the quickstarts before 0.24 is published ...

gsmet: It's not an issue. We need the latest master deployed to test the quickstarts anyway. You can just remove it.

using-mongodb-panache/pom.xml
loicmathieu: There is no test in the quickstart

using-mongodb-panache/src/main/resources/META-INF/resources/index.html
loicmathieu: done

loicmathieu:
@gsmet I renammed to `mongodb-panache`

using-mongodb-panache/pom.xml
loicmathieu: done

using-mongodb-panache/pom.xml
gsmet: Ah ah, silly me :).

loicmathieu:
@gsmet, I'll update the doc. I'll send a PR in a few minutes :)

cescoffier:
It would be great to have a few tests. I tend to use the quickstart as a good sensor to know when we break things.

loicmathieu:
@cescoffier as the quickstart uses the Quarkus __current__ version (0.23.2 as of today), it didn't really works untill the current master is released. So I cannot add tests before 0.24. 
You can create an issue and assign it to me, I'll take care of it after 0.24 will be released.

cescoffier:
The PR must target the `development` branch (which you do) and use the version 999-SNAPSHOT.
During the release, we have a script updating the versions to the released version and pushing to the `master` branch,

loicmathieu:
@cescoffier I just remember why I didn't include any test, if I keep the `999-SNAPSHOT` version on the pom I have the following issue (the same arise when I launch existing test like the one from the hibernate panache quickstat)

```
org.junit.jupiter.api.extension.TestInstantiationException: TestInstanceFactory [io.quarkus.test.junit.QuarkusTestExtension] failed to instantiate test class [org.acme.hibernate.orm.FruitsEndpointTest]: Failed to create the boostrap class loader
Caused by: java.lang.IllegalStateException: Failed to create the boostrap class loader
Caused by: io.quarkus.bootstrap.BootstrapException: Failed to create the deployment classloader for org.acme:hibernate-orm-resteasy::jar:1.0-SNAPSHOT
Caused by: io.quarkus.bootstrap.resolver.AppModelResolverException: Failed to read descriptor of org.acme:hibernate-orm-resteasy:jar:1.0-SNAPSHOT
Caused by: org.eclipse.aether.resolution.ArtifactDescriptorException: Failed to read artifact descriptor for org.acme:hibernate-orm-resteasy:jar:1.0-SNAPSHOT
Caused by: org.apache.maven.model.resolution.UnresolvableModelException: Could not find artifact io.quarkus:quarkus-bom:pom:999-SNAPSHOT
Caused by: org.eclipse.aether.resolution.ArtifactResolutionException: Could not find artifact io.quarkus:quarkus-bom:pom:999-SNAPSHOT
Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact io.quarkus:quarkus-bom:pom:999-SNAPSHOT
```

So I need to use a non-snapshot version to build the test (like 0.23.2), when it is ready, switch back to the 999-SNAPSHOT version and commit ...

I'll work on it later, I'll be on vacation for a week :sun_with_face:

gsmet:
OK, I'm merging this as we have the doc pointing to it. Let's add tests later.

cescoffier:
The issue is that it does not seem to work.

PR #294, opened 2019-09-26
Update the Keycloak quickstart to the new security support

cescoffier:


using-keycloak/src/main/java/org/acme/keycloak/AdminResource.java
```diff
@@ -25,6 +27,7 @@
  * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
  */
 @Path("/api/admin")
+@Authenticated
```
stuartwdouglas: This has not actually been wired up yet, it should be done before the next release though

cescoffier: The question is: should it be there, or `@RoleAllowed` should be enough.

I believe `@RoleAllowed` should be enough. WDYT?

using-keycloak/src/main/java/org/acme/keycloak/UsersResource.java
```diff
@@ -15,42 +15,40 @@
...
     }
 
     public class User {
 
         private final String userName;
 
-        User(KeycloakSecurityContext securityContext) {
-            this.userName = securityContext.getToken().getPreferredUsername();
+        User(SecurityIdentity identity) {
+            this.userName = identity.getPrincipal().getName();
```
sberyozkin: Is the access to the token has been lost ?

stuartwdouglas: In this case the Principal is an instance of JsonWebToken

using-keycloak/src/main/java/org/acme/keycloak/AdminResource.java
stuartwdouglas: @RolesAllowed("*") is the same

PR #298, opened 2019-10-01
kafka-quickstart - Remove the previous workaround that's no longer necessary

jaikiran:
Now that the upstream master branch of quarkus has an enhancement to deal with server side events better https://github.com/quarkusio/quarkus/pull/4167, the previous workaround that was added to this quickstart guide (https://github.com/quarkusio/quarkus/issues/2176) is no longer needed.
The commit here removes that workaround and uses the new construct for SSE. Manual testing with this change has gone fine.


cescoffier:
Thanks!

PR #301, opened 2019-10-02
Use the new HTTP request objects

cescoffier:
The guide must also be updated.


cescoffier:
The guide has been updated.

PR #304, opened 2019-10-03
feat: added rest-client example with Multipart

gastaldi:
This introduces a REST Client example using Multipart


cescoffier:
Thanks, should we have a note in the guide about this?

gastaldi:
Yeap, working on it

PR #313, opened 2019-10-09
Add ObjectStore to .gitignore

yrodiere:
Like we already do in the `quarkus` repository.

Because some people seem to be using Mac OS... I won't point fingers.


gsmet:
Thanks!

PR #314, opened 2019-10-09
Upgrade Hibernate Search to 6.0.0.Beta1

yrodiere:
@gsmet 

Necessary after quarkusio/quarkus#4478

I also added a small change to how the predicate is built, following our discussion the other day. Feel free to remove that change if you think it's too complex.


gsmet:
Merged, thanks!

PR #326, opened 2019-10-21
Remove unused property and test dependency

### LINKED ISSUE #325: Tidy up JWT quickstart
@sberyozkin could you have a look at these issues please?

 * https://github.com/quarkusio/quarkus-quickstarts/blob/development/using-jwt-rbac/pom.xml#L10 - looks like the property is not used anymore
 * https://github.com/quarkusio/quarkus-quickstarts/blob/development/using-jwt-rbac/pom.xml#L51 - we already have a dependency to JSON-B, could we use JSON-B instead of databind? Or is it not possible?

Once this is done, the guide will need an update. See this PR for where things are not OK: https://github.com/quarkusio/quarkus/pull/4548 .

sberyozkin:
Fixes #325

### DIFF
--- a/using-jwt-rbac/pom.xml
+++ b/using-jwt-rbac/pom.xml
@@ -7,7 +7,6 @@
   <version>1.0-SNAPSHOT</version>
   <properties>
     <surefire-plugin.version>2.22.1</surefire-plugin.version>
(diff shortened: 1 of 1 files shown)


gsmet:
Thanks!

PR #332, opened 2019-10-23
fix the network name + add information about producer restart

lhauspie:


gsmet:
@gunnarmorling can you have a look at that one?

kafka-streams-quickstart/README.md
```diff
@@ -32,14 +32,20 @@ A Docker Compose file is provided for running all the components.
 Start all containers by running
 
 ```bash
-docker-compose up --build
+docker-compose up -d --build
+```
+
+Due to an [an issue](https://github.com/smallrye/smallrye-reactive-messaging/issues/128) in the SmallRye Reactive Messaging component it might be that the producer application isn’t sending events if it was started before Apache Kafka was up. In this case, simply restart the producer (run from another shell window, keeping the other services running):
```
gsmet:
```suggestion
Due to [an issue](https://github.com/smallrye/smallrye-reactive-messaging/issues/128) in the SmallRye Reactive Messaging component it might be that the producer application isn’t sending events if it was started before Apache Kafka was up. In this case, simply restart the producer (run from another shell window, keeping the other services running):
```

gunnarmorling: The SmallRye issue has been resolved some time ago, so actually the entire note can be removed.

kafka-streams-quickstart/README.md
```diff
@@ -32,14 +32,20 @@ A Docker Compose file is provided for running all the components.
...
+```bash
+docker-compose stop producer && docker-compose start producer
 ```
 
 Now run an instance of the _debezium/tooling_ image which comes with several useful tools such as _kafkacat_ and _httpie_:
 
 ```bash
 docker run --tty --rm -i \
-        --network kafka-streams-quickstart_default \
+        --network kafkastreamsquickstart_default \
```
gunnarmorling: That's interesting, I could swear I've ran this many times. Has the default network name format changed perhaps? Eventually it would be better to use a fixed (and shorter) name in the Compose file, avoiding any confusion.

kafka-streams-quickstart/README.md
lhauspie: Oki, but I encountered the problem on my computer and I found this [Quarkus guide about kafak-streams](https://quarkus.io/guides/kafka-streams-guide) that mentions the producer restart.

> You should see log statements from the producer application about messages being sent to the "temperature-values" topic. Due to an [an issue](https://github.com/smallrye/smallrye-reactive-messaging/issues/128) in the SmallRye Reactive Messaging component it might be that the producer application isn’t sending events if it was started before Apache Kafka was up. In this case, simply restart the producer (run from another shell window, keeping the other services running):

kafka-streams-quickstart/README.md
lhauspie: The [kafak-streams guide](https://quarkus.io/guides/kafka-streams-guide) is using `docker run --tty --rm -i --network ks debezium/tooling:1.0`

So maybe we can armonize both by using the same one in this quistart repo ?

lhauspie: fixed

lhauspie: Ok, now I updated my docker-compose version from 1.16.1 to 1.24.1, the network has the right name (with dashes).
So fixing the network name is really a good idea to be consistent.

gsmet:
@gunnarmorling I let you decide what we should do with this one. I don't have a strong opinion on this. Thanks.

gunnarmorling:
+1 on the fixed network name.

On that warning sentence, this was there before and then was removed in
https://github.com/quarkusio/quarkus-quickstarts/commit/7b9b53833ae701fc338c6df54c9696405feff334#diff-f1aec0ddb30b65964e46dba363b9e95e.
So instead of re-adding it, it actually also should be removed from the
website (seems this was left there by accident). The referenced SR issue
has been resolved. If this actually still *is* an issue (I didn't observe
it any longer), the issue in SR should be re-opened / filed again to
address it.

lhauspie:
Oki.
I'll delete the warning sentence from readme.
I'll make a PR for removing the warning sentence from the website too.

amqp-quickstart/pom.xml
```diff
@@ -8,7 +8,7 @@
   <properties>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <surefire-plugin.version>2.22.0</surefire-plugin.version>
-    <quarkus.version>999-SNAPSHOT</quarkus.version>
+    <quarkus.version>0.26.1</quarkus.version>
```
cescoffier: Something went wrong, you must target 999-SNAPSHOT

cescoffier:
OH you need to rebase against the new development branch. The "BIG RENAMING" happened.

cescoffier:
Actually, let me bring this one home. I will just take the docker compose changes.

PR #344, opened 2019-10-30
Fixing the JWT check

sberyozkin:


sberyozkin:
It is never null itself as it is request scoped, so it was returning `hasJWT: true` with no tokens

sberyozkin:
OK

sberyozkin:
@cescoffier a couple of tests just receiving some numbers so I'm just checking if the responses are not empty

PR #346, opened 2019-10-31
Fix mismatch between the guide and the quickstart.

cescoffier:
The difference was introduced because of the tests.
Actually the tests were not using Kafka at all.
This commit starts Kafka as part of the tests and avoids having the configuration mismatch.


gsmet:
Well, I trust you on this and we'll see if CI passes.

PR #349, opened 2019-11-02
add cron jobs in scheduler quickstart

machi1990:
Following https://github.com/quarkusio/quarkus/pull/5127
/cc @gsmet


scheduler-quickstart/src/main/java/org/acme/scheduling/CounterBean.java
```diff
@@ -20,4 +21,13 @@ void increment() {
         counter.incrementAndGet();
     }
     
-}
\ No newline at end of file
+    @Scheduled(cron="0 15 10 * * ?")
```
cescoffier: Would it be possible to make it testable and extend the tests?

scheduler-quickstart/src/main/java/org/acme/scheduling/CounterBean.java
```diff
@@ -20,4 +21,13 @@ void increment() {
...
+    void cronJob(ScheduledExecution execution) {
+        System.out.println(execution.getScheduledFireTime());
+    }
+
+    @Scheduled(cron = "{cron.expr}")
+    void cronJobWithExpressionInMpConfig() {
```
cescoffier: InConfig, instead of inMpConfig

machi1990:
> Was the guide also updated?

Yes, the guide has been updated here https://github.com/quarkusio/quarkus/pull/5127

scheduler-quickstart/src/main/java/org/acme/scheduling/CounterBean.java
machi1990: I am thinking about incrementing the counter for each of the two added methods. Is that acceptable?

cescoffier: +1

machi1990: Done, I have updated the two methods with a call that increment the counter. 
Same modification applied in the guide.

gsmet:
Force pushed to fix a minor indentation issue and merged.

PR #409, opened 2019-12-27
since vert.x 3.8.0,we should use request instead of send.

lizhuoxuan:
method "send" will be remove in version 4,now we should use
`eventBus.request("the-address", body, ar -> ...);`
here is the wiki [https://github.com/vert-x3/wiki/wiki/3.8.0-Deprecations-and-breaking-changes](url "Vert.x wiki")


gsmet:
@cescoffier @vietj can you check that one? Thanks!

PR #420, opened 2020-01-15
Implement a periodic workflow verifying the QS against the Quarkus master

cescoffier:
The implemented workflow runs every 6 hours and verifies the quickstarts against the latest Quarkus (master branch). It builds the native executables using the container policy and runs native ITs.

To ease debugging and to avoid waiting for the next _tick_, the dispatch event is also enabled. So we can trigger the workflow by issuing a curl command as follows:

```
curl -X POST -H "Authorization: token $GITHUB_TOKEN" \
          -H "Accept: application/vnd.github.everest-preview+json"  \
         -H "Content-Type: application/json" \
          https://api.github.com/repos/quarkusio/quarkus-quickstarts/dispatches \
         --data '{"event_type": "internal_event"}'
```

The build reporting uses Github Issues. It creates, closes or adds a comment to an issue named `Quickstarts Native Build failed against Quarkus Master`. If the issue does not exist it creates it at the first failure. It the issue exist and the build passes, the issue is closed (with a comment). If the build still fails, a comment is added. The comment contains the SHAs of the quickstart and quarkus commits.

At the moment the STM quickstart is ignored. An issue has been found yesterday and is currently being reviewed (https://github.com/quarkusio/quarkus/issues/6553). 

**IMPORTANT**: The build takes almost 3h to run...

### DIFF
--- a/.github/NativeBuildReport.kts
+++ b/.github/NativeBuildReport.kts
@@ -21,34 +21,36 @@ if (status == "cancelled") {
     System.exit(0)
 }
 
-val ISSUE_TITLE = "Quickstarts Native Build failed against Quarkus Master"
+val ISSUE_TITLE = "[CI] - Quickstarts Native Build failed against Quarkus Master"
 val REPO = "quarkusio/quarkus-quickstarts"
 
 val github = GitHubBuilder().withOAuthToken(token).build()
-val repository = github.getRepository("quarkusio/quarkus-quickstarts")
-val issues = repository.getIssues(GHIssueState.OPEN)
-val existingIssue = issues.firstOrNull { i -> i.getTitle() == ISSUE_TITLE}
+val repository = github.getRepository("quarkusio/quarkus")
+val issues = repository.getIssues(GHIssueState.ALL)
+
+val issue = issues.firstOrNull { i -> i.getTitle() == ISSUE_TITLE}
 
 val quickstartsCommit = getRepositoryCommit(".")
 val quarkusCommit = getRepositoryCommit("quarkus")
 if (succeed) {
-    if (existingIssue != null) {
+    if (issue != null  && isOpen(issue)) {
         // close issue with a comment
-        val comment = existingIssue.comment("""
+        val comment = issue.comment("""
             Build fixed with:
 
             * Quarkus commit: ${quarkusCommit}
-            * Quickstarts commit: ${quickstartsCommit}               
+            * Quickstarts commit: ${quickstartsCommit}  
+            * Link to build: https://github.com/quarkusio/quarkus-quickstarts/actions
 
         """.trimIndent())        
-        existingIssue.close()
-        println("Comment added on issue ${existingIssue.getHtmlUrl()} - ${comment.getHtmlUrl()}, the issue has also been closed")
+        issue.close()
+        println("Comment added on issue ${issue.getHtmlUrl()} - ${comment.getHtmlUrl()}, the issue has also been closed")
     } else {
-        println("Nothing to do - the build passed and there are no issue to close")
+        println("Nothing to do - the build passed and the issue is already closed or not created")
     }
 } else  {
     // Build failed
-    if (existingIssue == null) {
+    if (issue == null) {
         // create a new issue
         val newIssue = repository.createIssue(ISSUE_TITLE)
             .body("""
@@ -58,21 +60,41 @@ if (succeed) {
                 The build failed with:
 
                 * Quarkus commit: ${quarkusCommit}
-                * Quickstarts commit: ${quickstartsCommit}               
+                * Quickstarts commit: ${quickstartsCommit}     
+                * Link to build: https://github.com/quarkusio/quarkus-quickstarts/actions 
+                         
+                This issue will be closed when the build gets fixed. It will be re-opened if it fails again later.
+                Do not close the issue yourself (it will be re-opened automatically)
+                You can subscribe and monitor the status.
 
             """.trimIndent())
-            .create()
+                .label("area/infra")
+                .create()
         println("New issue created: ${newIssue.getTitle()} - ${newIssue.getHtmlUrl()}")    
     } else {
-        // comment on issue
-        val comment = existingIssue.comment("""
+        if (isOpen(issue)) {
+            val comment = issue.comment("""
             Build still failing with:
 
             * Quarkus commit: ${quarkusCommit}
             * Quickstarts commit: ${quickstartsCommit}   
+            * Link to build: https://github.com/quarkusio/quarkus-quickstarts/actions
 
         """.trimIndent())
-        println("Comment added on issue ${existingIssue.getHtmlUrl()} - ${comment.getHtmlUrl()}")
+            println("Comment added on issue ${issue.getHtmlUrl()} - ${comment.getHtmlUrl()}")
+        } else {
+            issue.reopen()
+            val comment = issue.comment("""
+            Unfortunately, the build fails again:
+
+            * Quarkus commit: ${quarkusCommit}
+            * Quickstarts commit: ${quickstartsCommit}   
+            * Link to build: https://github.com/quarkusio/quarkus-quickstarts/actions
+
+        """.trimIndent())
+            println("Comment added on issue ${issue.getHtmlUrl()} - ${comment.getHtmlUrl()}, the issue has been re-opened.")
+
+        }
 
     }
 }
@@ -97,5 +119,9 @@ fun String.runCommand(workingDir: String): String {
         e.printStackTrace()
         return ""
     }
+}
+
+fun isOpen(issue : GHIssue) : Boolean {
+    return issue.getState() == GHIssueState.OPEN
 }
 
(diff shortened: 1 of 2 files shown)


cescoffier:
Forgot to mention that the workflow uses Java 8 as it does the native build.

cescoffier:
@emmanuelbernard ^

.github/workflows/native-build-development.yml
```diff
@@ -0,0 +1,57 @@
...
+          && cd quarkus \
+          && mvn -B clean install -DskipTests -DskipITs
+
+      - name: Build Quickstart with native
+        run: |
+          mvn -B clean install -Pnative \
+            -Dquarkus.native.container-build=true \
+            -pl "!software-transactional-memory-quickstart"
+
+      - name: Report Success
```
cescoffier: Might need to remove "success" as it reports failure too

.github/NativeBuildReport.kts
```diff
@@ -0,0 +1,101 @@
...
+if (status == "cancelled") {
+    println("Job status is `cancelled` - exiting")
+    System.exit(0)
+}
+
+val ISSUE_TITLE = "Quickstarts Native Build failed against Quarkus Master"
+val REPO = "quarkusio/quarkus-quickstarts"
+
+val github = GitHubBuilder().withOAuthToken(token).build()
+val repository = github.getRepository("quarkusio/quarkus-quickstarts")
```
emmanuelbernard: I think we should open it on quarkusio/quarkus because I am afraid not enough people have eyeballs on the quarkus-quickstart project issues

.github/NativeBuildReport.kts
```diff
@@ -0,0 +1,101 @@
...
+val issues = repository.getIssues(GHIssueState.OPEN)
```
emmanuelbernard: I wonder for the sake of simplicity if we should not have one issue that is closed and reopened when the quickstarts fail. My main (only reason) is that we can enforce that all the relevant persons are subscribed to the issue and they will get notification through GitHub. This will be more visible than a random new issue created.

Alternatively, we have a list of people we notify (I'd like to be a part of).

.github/NativeBuildReport.kts
```diff
@@ -0,0 +1,101 @@
...
+        // create a new issue
+        val newIssue = repository.createIssue(ISSUE_TITLE)
+            .body("""
+                The build of the  `development` branch of the quickstarts against quarkus `master` has failed.
+                This build verifies the native compilation and executes the native integration tests.
+
+                The build failed with:
+
+                * Quarkus commit: ${quarkusCommit}
+                * Quickstarts commit: ${quickstartsCommit}
```
emmanuelbernard: is it possible to point to the actual error like a jenkins-like report. I am not sure GitHub action has such facility so I'm asking. I'm trying to see what will be the first step of the person finding the regression, likely look at the error report to trim down the cause. If not the user would have to get the quickstarts and run quarkus-core and the quickstart tests locally.

.github/NativeBuildReport.kts
```diff
@@ -0,0 +1,101 @@
...
+fun String.runCommand(workingDir: String): String {
+    try {
+        val parts = this.split("\\s".toRegex())
+        val proc = ProcessBuilder(*parts.toTypedArray())
+                .directory(File(workingDir))
+                .redirectOutput(ProcessBuilder.Redirect.PIPE)
+                .redirectError(ProcessBuilder.Redirect.PIPE)
+                .start()
+
+        proc.waitFor(60, TimeUnit.MINUTES)
```
emmanuelbernard: 60 mins for `git rev-parse HEAD`, you have no faith in computers ;)

.github/NativeBuildReport.kts
cescoffier: Yes, it should work. I need to check that the token I used has access to this repo.

.github/NativeBuildReport.kts
cescoffier: We can do this, and use a single issue. We just need to be sure that no-one will ever create another issue with the same title.

.github/NativeBuildReport.kts
cescoffier: unfortunately, it's where things get hairy... I didn't find a way to get the log, neither download it. I will have another look.

.github/NativeBuildReport.kts
cescoffier: Oh... my faith is gone a long time ago... That's why I do reactive stuff :-D

.github/workflows/native-build-development.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+          git clone https://github.com/quarkusio/quarkus.git \
+          && cd quarkus \
+          && mvn -B clean install -DskipTests -DskipITs
+
+      - name: Build Quickstart with native
+        run: |
+          mvn -B clean install -Pnative \
+            -Dquarkus.native.container-build=true \
+            -pl "!software-transactional-memory-quickstart"
```
geoand: Is this the one that has a bug for the time being?

.github/workflows/native-build-development.yml
geoand: Makes sense

.github/NativeBuildReport.kts
geoand: I like this idea, since subscriptions would just work as opposed to have the relevant constantly monitor for new issues

.github/workflows/native-build-development.yml
cescoffier: yes (STM)

.github/NativeBuildReport.kts
cescoffier: ok, there is no way to do it for now... Even linking to the execution is not reliable.

.github/workflows/native-build-development.yml
geoand: Thanks

.github/NativeBuildReport.kts
cescoffier: I can definitely add [CI] and add the link to the action page (see next comment about the impossibility to retrieve the log from inside the workflow)

geoand: Sounds good to me

.github/workflows/native-build-development.yml
cescoffier: Fixed.

.github/NativeBuildReport.kts
cescoffier: Should work with the current token.

.github/NativeBuildReport.kts
cescoffier: New title: `[CI] - Quickstarts Native Build failed against Quarkus Master` label `area/infra`.

.github/NativeBuildReport.kts
cescoffier: Added a plain link to the actions page.

cescoffier:
There is one thing that worries me...
The issue lookup is taking a lot of time because the main quarkus repo has a lot of issues (opened and closed). I was wondering if I should not create the issue manually, and do a lookup by id which would be much faster. 

WDYT?

geoand:
> There is one thing that worries me...
> The issue lookup is taking a lot of time because the main quarkus repo has a lot of issues (opened and closed). I was wondering if I should not create the issue manually, and do a lookup by id which would be much faster.
> 
> WDYT?

I agree, let's just do that instead of risk something timing out and us loosing the report completely

cescoffier:
Also, with https://github.com/quarkusio/quarkus/pull/6555, we can remove the workaround for STM.

.github/NativeBuildReport.kts
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env kscript
+
+@file:MavenRepository("jcenter","https://jcenter.bintray.com/")
+@file:MavenRepository("maven-central","https://repo1.maven.org/maven2/")
```
gsmet: I think we should use `https://repo.maven.apache.org/maven2/` instead.

.github/workflows/native-build-development.yml
gsmet: I merged the PR that is supposed to fix it so we should be able to include it now.

.github/NativeBuildReport.kts
cescoffier: I actually never used that one, thanks!

cescoffier:
Alright, it's now going to report on https://github.com/quarkusio/quarkus/issues/6588.

cescoffier:
The CI failed... this is exactly why we need this in because it would report this issue in quarkus.

geoand:
I will probably copy the general idea of this in order to setup a periodic CI run of the Quarkus Platform (when the pieces are in place which they aren't yet but should be soon).

geoand:
@cescoffier the quickstart issues should be fixed now, can you rebase the PR?

cescoffier:
@geoand done.

geoand:
@cescoffier We aren't waiting for anything else for this one right? Should we merge and have it start working?

cescoffier:
Nothing I’m aware of.

geoand:
Merging then

PR #424, opened 2020-01-17
Handle association manually as automatic association management was disabled

gsmet:
See https://github.com/quarkusio/quarkus/pull/6569

/cc @geoand @cescoffier


geoand:
@gsmet the FT one now breaks (not the fault of this PR of course). 
Since this test was actually tested (due to the HS quickstart being run before the FT one), I propose we merge this in order to have one less problem to deal with

PR #426, opened 2020-01-17
Simplify build pipeline

geoand:


gsmet:
I'm not sure about this. @cescoffier had good reasons to use the external action as it installs OpenJDK and IIRC GitHub installs the Azul one.

geoand:
Then I'd definitely like to hear them :)

gsmet:
well, see the end of my sentence :)

We want our things to be tested with OpenJDK not the Azul one.

geoand:
> well, see the end of my sentence :)
> 
> We want our things to be tested with OpenJDK not the Azul one.

That's it? I thought there was something more than that :P

IMHO I don't think that is a good enough reason to not use the simplified way of doing things. Azul is just a build of OpenJDK in any case

gsmet:
Is there a way to use a matrix to make the JDK names variables? That would already simplify things while still using official OpenJDK builds.

geoand:
I think so. Let me check

cescoffier:
We want to test against openjdk, which is the most used. If you want azul, add another job.

And, yes, we could you matrix, bit didn’t find them very useful.

geoand:
Let's see if my latest test works

geoand:
I think this latest version I added makes sense

cescoffier:
Ok for now, but it also avoids having the possibility to test with another impl like OpenJ9.

geoand:
Sure, but we can always cross that boat when we get there :)

PR #429, opened 2020-01-17
Use GraalVM 19.3.1 when checking native image build in CI

geoand:


cescoffier:
We need to remember to update it.

geoand:
Yes absolutely

PR #432, opened 2020-01-20
add trace to debug the status issue

cescoffier:
This is about https://github.com/quarkusio/quarkus/issues/6588#issuecomment-576046728 - trying to see what kind of status gets the step.


cescoffier:
This commit will need to be reverted once the issue is fixed.

PR #434, opened 2020-01-22
Capture rss of quarkus application. Fail build if RSS is greater than…

johnaohara:
… 25MB

This PR adds an RSS check to the build_with_native workflow.  A threshold of 25MB has been set, and if the getting-started quickstart uses more memory than the threshold, the build fails.

I don't like inserting the `sleep 2` command to wait for the background process to start, but in my testing, I could not find a reliable way to grep for the startup message from a process running in the background running on the GH actions CI.

### DIFF
--- a/.github/workflows/native-build-development.yml
+++ b/.github/workflows/native-build-development.yml
@@ -29,9 +29,9 @@ jobs:
 
       - name: Build Quarkus master
         run: |
-          git clone https://github.com/quarkusio/quarkus.git \
-          && cd quarkus \
-          && mvn -B clean install -DskipTests -DskipITs
+          git clone https://github.com/quarkusio/quarkus.git
+          cd quarkus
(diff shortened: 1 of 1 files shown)


geoand:
Thanks @johnaohara !

@cescoffier is it fine with you?

johnaohara:
Perhaps hold off merging this for the moment, noticed that we want to capture start time as well

johnaohara:
Have added a startup threshold as well

geoand:
+1 from me

cescoffier:
I've approved the change. let me know when it all ok for you.

I was wondering if we want to collect the measure somewhere.

geoand:
So @johnaohara @cescoffier are we going to get this one in now and potentially improve later?

cescoffier:
+1 

@gsmet is it safe to merge now?

PR #437, opened 2020-01-23
Fix Kafka Quickstart

stuartwdouglas:


stuartwdouglas:
This pattern only worked if there was a single test class in the test suite, which is ok for quickstarts but not going to be workable for real apps. As soon as you add a second test class this would not work, as you need the resource to be scoped to the entire run, not a single class.

cescoffier:
Thanks @stuartwdouglas. That makes sense. We would just need to extend the testing doc about this (maybe it's already done, didn't check).

PR #441, opened 2020-01-27
Create native "getting started" image without having to install GraalVM

SchulteMarkus:
"-Dquarkus.native.container-build=true" can be added to quarkus-maven-plugin for creating a native "getting started" image without the need to install GraalVM locally. Instead, docker image quay.io/quarkus/ubi-quarkus-native-image is used for as build image. 
See https://quarkus.io/guides/building-native-image#quarkus-native-pkg-native-config_quarkus.native.container-build for details


getting-started/README.md
```diff
@@ -72,7 +72,7 @@ Compiling a native executable takes a bit longer, as GraalVM performs additional
 steps to remove unnecessary codepaths. Use the  `native` profile to compile a
 native executable:
 
-> ./mvnw package -Dnative
+> ./mvnw package -Dnative -Dquarkus.native.container-build=true
```
cescoffier: The flag is only required to build a container-compatible executable.

gsmet:
We require GraalVM at the beginning of the README so I'm not sure about that.

In any case, the PR should target the `development` branch or it will get erased.

SchulteMarkus:
Closing here as wrong base branch.

SchulteMarkus:
Follow-up is #442

getting-started/README.md
SchulteMarkus: You seem to be wrong here. Omitting the flag requires ```native-image``` being present as executable locally. Comment please in follow-up #442

SchulteMarkus:
> We require GraalVM at the beginning of the README so I'm not sure about that.
> 
> In any case, the PR should target the `development` branch or it will get erased.

PR recreated with correct base branch at #442

Issue #433, opened 2020-01-20
QS are using quarkus-bom and not quarkus-universe-bom


rsvoboda:
QS are using quarkus-bom and not quarkus-universe-bom.

quarkus-universe-bom is the default for new project generated by `io.quarkus:quarkus-maven-plugin:create`

I can see benefit of quarkus-bom approach on development branch when running against quarkus master. But what about released versions of QS / master ? Should it be using quarkus-universe-bom or not ?

Side info about propertiesdefined in pom.xml:
Properties from config-quickstart
```xml
  <properties>
    <quarkus.version>1.1.1.Final</quarkus.version>
    <surefire-plugin.version>2.22.1</surefire-plugin.version>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <maven.compiler.source>1.8</maven.compiler.source>
    <maven.compiler.target>1.8</maven.compiler.target>
  </properties>
```
Properties from io.quarkus:quarkus-maven-plugin:create based project
```xml
  <properties>
    <compiler-plugin.version>3.8.1</compiler-plugin.version>
    <maven.compiler.parameters>true</maven.compiler.parameters>
    <maven.compiler.source>1.8</maven.compiler.source>
    <maven.compiler.target>1.8</maven.compiler.target>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
    <quarkus-plugin.version>1.1.1.Final</quarkus-plugin.version>
    <quarkus.platform.artifact-id>quarkus-universe-bom</quarkus.platform.artifact-id>
    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
    <quarkus.platform.version>1.1.1.Final</quarkus.platform.version>
    <surefire-plugin.version>2.22.1</surefire-plugin.version>
  </properties>
```

gsmet:
Please do not fix that right away as it's a much more complicated issue than it looks.

rsvoboda:
I do not plan to do anything about this. I want discussion about this first. As I wrote before I see benefits of current approach for development branch. Doing switch for master/tags is non trivial task and I don't take quarkus-platform complexity into account.

maxandersen:
whats the challenges to have quickstart use latest ? we should be able to use 999-snapshot of platform too?

gsmet:
[...]

PR #447, opened 2020-02-03
Change platform version to 1.2.0.Final

rsvoboda:
Change platform version to 1.2.0.Final for spring-security-quickstart

Targets `master` on purpose

@gsmet / @cescoffier release procedure needs to be adjusted to cover version rewrite of Quarkus platform property


gsmet:
The right fix is to not use the platform in the quickstarts.

rsvoboda:
I can do that for development branch.

Is it needed for master too ?

gsmet:
Please do it for the `development` branch, I'll cherry-pick (and fix the conflicts) in master.

maxandersen:
> The right fix is to not use the platform in the quickstarts.

whats the reason why we should not use platform in in the quickstarts ? aren't they supposed to be used as examples on how to write apps for the quarkus platform as opposed to quarkus core ?

rsvoboda:
@maxandersen https://github.com/quarkusio/quarkus-quickstarts/issues/433 fyi

PR #452, opened 2020-02-05
add missing sasl-mechanism property

karesti:
Latest infinispan server image won't work because auth is enabled and the sasl mechanism needs to be set up


cescoffier:
Thanks!

PR #457, opened 2020-02-06
validate/format on PR's

maxandersen:
depends on quarkusio/quarkus#7015 being merged and then we'll need to figure out what version of the ide-plugin to point to, atm. its 999-SNAPSHOT.

**Check list**:

Your pull request:

- [ ] makes sure the documentation must not be updated

### DIFF
--- a/vertx-quickstart/src/test/java/org/acme/vertx/FruitsEndpointTest.java
+++ b/vertx-quickstart/src/test/java/org/acme/vertx/FruitsEndpointTest.java
@@ -31,29 +31,32 @@ public class FruitsEndpointTest {
     public void testListAllFruits() {
         //List all, should have all 3 fruits the database has initially:
         given()
-                .when().get("/fruits")
-                .then()
-                .statusCode(200)
-                .body(
-                        containsString("Orange"),
-                        containsString("Pear"),
-                        containsString("Apple"));
+               .when()
+               .get("/fruits")
+               .then()
+               .statusCode(200)
+               .body(
+                       containsString("Orange"),
+                       containsString("Pear"),
+                       containsString("Apple"));
 
         //Delete the Orange:
         given()
-                .when().delete("/fruits/1")
-                .then()
-                .statusCode(204);
+               .when()
+               .delete("/fruits/1")
+               .then()
+               .statusCode(204);
 
         //List all, Orange should be missing now:
         given()
-                .when().get("/fruits")
-                .then()
-                .statusCode(200)
-                .body(
-                        not(containsString("Orange")),
-                        containsString("Pear"),
-                        containsString("Apple"));
+               .when()
+               .get("/fruits")
+               .then()
+               .statusCode(200)
+               .body(
+                       not(containsString("Orange")),
+                       containsString("Pear"),
+                       containsString("Apple"));
     }
 
 }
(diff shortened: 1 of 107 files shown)


rsvoboda (changes requested):
You need to have Check Formatting step after Build Quarkus master because of 999-SNAPSHOT version being used

maxandersen:
> You need to have Check Formatting step after Build Quarkus master because of 999-SNAPSHOT version being used

good point - updated.

maxandersen:
the CI wont pass as the code wasn't yet formatted - so I've rebased on latest development, added a `format-quickstarts` command and ran it to get the right formatting.

recommedning to push this ASAP before others get caught in using old formatted files ;)

vertx-quickstart/src/test/java/org/acme/vertx/FruitsEndpointTest.java
```diff
@@ -20,40 +20,24 @@
...
+        // List all, should have all 3 fruits the database has initially:
+        given().when().get("/fruits").then().statusCode(200).body(containsString("Orange"), containsString("Pear"),
+                containsString("Apple"));
+
+        // Delete the Orange:
+        given().when().delete("/fruits/1").then().statusCode(204);
+
+        // List all, Orange should be missing now:
+        given().when().get("/fruits").then().statusCode(200).body(not(containsString("Orange")), containsString("Pear"),
+                containsString("Apple"));
```
gsmet: Frankly, this is not an improvement.

The tests are made to be fully understandable and this definitely doesn't help understanding what's going on.

@cescoffier WDYT?

gsmet (changes requested):

vertx-quickstart/src/test/java/org/acme/vertx/FruitsEndpointTest.java
maxandersen: i didn’t do anything special here - just used quarkus default formatting style. 

I can try and tweak the format using https://stackoverflow.com/questions/18324017/how-to-set-eclipse-code-formatter-to-support-fluent-interfaces or we can to be safe utilize
```
// @formatter:off
...
// @formatter:on
```

to have ful control

maxandersen: so the setting in SO works in eclipse giving me this:

```
   public void testListAllFruits() {
        // List all, should have all 3 fruits the database has initially:
        given().when()
               .get("/fruits")
               .then()
               .statusCode(200)
               .body(containsString("Orange"),
                       containsString("Pear"),
                       containsString("Apple"));

        // Delete the Orange:
        given().when()
               .delete("/fruits/1")
               .then()
               .statusCode(204);

        // List all, Orange should be missing now:
        given().when()
               .get("/fruits")
               .then()
               .statusCode(200)
               .body(
                       not(containsString("Orange")),
                       containsString("Pear"),
                       containsString("Apple"));
    }
```

but looks like https://code.revelc.net/formatter-maven-plugin does not support that setting ..at least it seem to ignore it ;/ . investigating.

maxandersen: hmm - seems like the formatter completely ignores the formatter when no config in the pom.xml....explains why its not formatting but sucks if true ;/

maxandersen: okey so there is unfortunately no way to easily have the formatter pick up the shared resources via mvn unless we add formatter settings to each plugin.

as work around i've made it so it points to the format via ../quarkus.

i've also committed code is it would look if change formatter to automatically put a linebreak and indent call chains per nested column.

This would require a change to the share format to work.

cescoffier:
@maxandersen / @gsmet What should we do about this one? 1944 days.... I guess we can close it.

PR #460, opened 2020-02-07
Eliminate wildcard imports in QS and added script for check QS

mirostary:
Eliminate wildcard imports in QS and added script for check QS to ensure no new appear in the codebase.

**Check list**:

Your pull request:

- [x] links the documentation update pull request (if needed)
           https://github.com/quarkusio/quarkus/pull/7068

### DIFF
--- a/.github/workflows/build-pr-development.yml
+++ b/.github/workflows/build-pr-development.yml
@@ -19,6 +19,13 @@ jobs:
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-
+      - name: Check Wildcards
+        run: |
+          if [ $(git grep import | grep '\*;' | grep -E -v '(.mvn/|README.md|CONTRIBUTING.md)' | wc -l) != 0 ]; then
+            echo "Please don't use star imports in your Java files: "
+            git grep import | grep '\*;' | grep -E -v '(.mvn/|README.md|CONTRIBUTING.md)'
+            exit 1
+          fi
       - name: Install JDK {{ matrix.java }}
         uses: joschi/setup-jdk@v1.0.0
         with:
--- a/mongodb-quickstart/src/main/java/org/acme/rest/json/ReactiveFruitResource.java
+++ b/mongodb-quickstart/src/main/java/org/acme/rest/json/ReactiveFruitResource.java
@@ -4,7 +4,11 @@ import java.util.List;
 import java.util.concurrent.CompletionStage;
 
 import javax.inject.Inject;
-import javax.ws.rs.*;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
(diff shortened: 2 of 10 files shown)


maxandersen:
why are we against .* imports here ? we seem to use it wildly everywhere else in quarkus when more than a few classes used from same package ?

rsvoboda:
Hi @maxandersen, I asked for this as I wanted smaller tasks for guys to start with and to get a bit familiar with quickstarts codebase.

General motivation for eliminating .* imports: 
 - https://rules.sonarsource.com/java/RSPEC-2208
 - https://medium.com/@aayushsharda/why-is-it-better-to-avoid-in-import-statements-in-java-ff9ceaca8aeb
I'm aware that wildcard imports topic can easily turn into discussion similar to "tabs vs. spaces".

> we seem to use it wildly everywhere else in quarkus

From my perspective Quickstarts are a bit different comparing to other quarkusio codebase.
Quickstarts are showcase for the end users and I'd rather be explicit about the imports.
Spring only users / non-jaxrs users looking into them could appreciate explicit import statement.

maxandersen:
I grok the want for reducing import .*'s, but static imports ? your links explicitly says they are excluded from this general convention.

ie. "import static io.restassured.RestAssured.*;" should be allowed to stay imo as that API is referring to a full class name. Adding each method manually is just overkill.

gsmet:
AFAIK, we are requiring not using start imports for everything. At least that's the rule we apply in Quarkus.

So this one looks OK. Maybe let's try it and see if it gets in the way too much and we can tune it if it's the case.

.github/workflows/build-pr-development.yml
```diff
@@ -19,6 +19,13 @@ jobs:
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-
+      - name: Check Wildcards
+        run: |
+          if [ $(git grep import | grep '\*;' | grep -E -v '(.mvn/|README.md|CONTRIBUTING.md)' | wc -l) != 0 ]; then
+            echo "Wildcard found!"
```
gsmet: This is not a helpful error message. `Please don't use star imports in your Java files:` would be better.

rsvoboda: PR was updated as requested, resolving the conversation

rsvoboda:
@gsmet your comment was addressed, ready for merge.

@mirostary please @mention requestor of the change so he gets notified. There are many PRs and it's impossible for gatekeeper to circle across all of them.

maxandersen:
> AFAIK, we are requiring not using start imports for everything. At least that's the rule we apply in Quarkus.

okey, weird to me and 19 instances of it in the quarkus code base :)
Having to explicitly import all hamcrest matcher methods for example gets quite noisy if not using .*.

maxandersen:
anyway, as long as there are tools to verify/clean it up I'm not against it - just find it silly to require to explicit import individually methods :)

gsmet:
Merged, thanks!

PR #465, opened 2020-02-18
Switch to KafkaRecord.

cescoffier:
KafkaMessage has been deprecated and is now replaced by KafkaRecord.


geoand:
Thanks for fixing so quickly @cescoffier and for checking while I was out @gsmet

Issue #353, opened 2019-11-04
Add tests to the neo4j quickstart


cescoffier:
* Add Quarkus tests
* Add native tests

PR #475, opened 2020-02-25
[#353] Add tests to the neo4j quickstart

mcserra:
Signed-off-by: Miguel Serra <serramiguel1@gmail.com>

### DIFF
--- a/neo4j-quickstart/pom.xml
+++ b/neo4j-quickstart/pom.xml
@@ -12,6 +12,7 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.target>1.8</maven.compiler.target>
+    <docker-plugin.version>0.28.0</docker-plugin.version>
   </properties>
   <dependencyManagement>
     <dependencies>
@@ -69,6 +70,58 @@
           <execution>
             <goals>
               <goal>build</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <!-- Automatically start Neo4J for integration testing - requires Docker -->
+        <groupId>io.fabric8</groupId>
+        <artifactId>docker-maven-plugin</artifactId>
+        <version>${docker-plugin.version}</version>
+        <configuration>
+          <skip>${skipTests}</skip>
+          <images>
+            <image>
+              <name>neo4j:4.0.0</name>
+              <alias>neo4j-database</alias>
+              <run>
+                <env>
+                  <NEO4J_AUTH>neo4j/secret</NEO4J_AUTH>
+                </env>
+                <ports>
+                  <port>7474:7474</port>
+                  <port>7687:7687</port>
+                </ports>
+                <log>
+                  <prefix>neo4j:</prefix>
+                  <date>default</date>
+                  <color>green</color>
+                </log>
+                <wait>
+                  <http>
+                    <url>http://0.0.0.0:7474/browser</url>
+                  </http>
+                  <time>20000</time>
+                </wait>
+              </run>
+            </image>
+          </images>
+        </configuration>
+        <executions>
+          <execution>
+            <id>docker-start</id>
+            <phase>test-compile</phase>
+            <goals>
+              <goal>stop</goal>
+              <goal>start</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>docker-stop</id>
+            <phase>post-integration-test</phase>
+            <goals>
+              <goal>stop</goal>
             </goals>
           </execution>
         </executions>
--- a/neo4j-quickstart/pom.xml
+++ b/neo4j-quickstart/pom.xml
@@ -12,7 +12,7 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <docker-plugin.version>0.28.0</docker-plugin.version>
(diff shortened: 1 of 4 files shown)


neo4j-quickstart/pom.xml
```diff
@@ -73,6 +74,58 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <!-- Automatically start Neo4J for integration testing - requires Docker -->
+        <groupId>io.fabric8</groupId>
```
cescoffier: Any reason to not use TestContainers as others quickstart (like Kafka)

neo4j-quickstart/src/test/java/org/acme/neo4j/FruitsEndpointTest.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.hamcrest.core.IsNot.not;
+
+@QuarkusTest
+class FruitsEndpointTest  {
+
+    private static final Map<String, String> fruitIdMap = new HashMap<>();
+    private static final UnaryOperator<String> FRUIT = fruit ->  String.format("{\"name\":\"%s\"}", fruit);
+
+    @Test
+    public void testResource() {
```
cescoffier: Would be great to also test the reactive API if possible.

neo4j-quickstart/pom.xml
mcserra: Not really. I thought it'd be better to folow the hibernate quickstarts. I do prefer testcontainers, and didn't noticed it was used. Do you want me to change it?

neo4j-quickstart/src/test/java/org/acme/neo4j/FruitsEndpointTest.java
mcserra: I use it in line 60. Would it be better to create other test class? It doesn't have a persist method, so I did it in the same one

cescoffier: sorry, I didn't see it! that's great!

neo4j-quickstart/pom.xml
cescoffier: The advantage of test container is that you can run the tests from your IDE or from your terminal.

mcserra: Ok, the changes are done. I don't think the changes can have something to do with the fail on the pipeline. Can we try and run it again?

neo4j-quickstart/src/test/java/org/acme/neo4j/Neo4jResource.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.testcontainers.containers.Neo4jContainer;
+
+public class Neo4jResource implements QuarkusTestResourceLifecycleManager {
+
+    private final Neo4jContainer NEO4J = new Neo4jContainer<>("neo4j:4.0.0").withAdminPassword(null)
+        .withExposedPorts(7687)
+        .withCreateContainerCmdModifier(cmd -> cmd
+            .withHostName("localhost")
+            .getHostConfig().withPortBindings(new PortBinding(Ports.Binding.bindPort(7687), new ExposedPort(7687))));
```
cescoffier: The interest of test container combined with this approach is that you don't need to set up the exposed port. It will pick a random available port.

cescoffier:
The CI failure is unrelated. Let's see on the next attempt.

neo4j-quickstart/src/test/java/org/acme/neo4j/Neo4jResource.java
mcserra: I had a big problem to make it work with the native test, and found that solution. But I just tried it again without that and it worked, so I must have made a mistake...

cescoffier: @mcserra it used to be the case, however look at how the other QS uses the QuarkusTestResource, and then it works :-)

mcserra: Its pretty cool :) Is it ready then?

neo4j-quickstart/pom.xml
cescoffier: Yes, look at https://github.com/quarkusio/quarkus-quickstarts/blob/development/kafka-quickstart/src/test/java/org/acme/kafka/KafkaResource.java for example.

mcserra: I saw it, and the change is already done in the last commit, I was just asking if there is any other change that needs to be made...

cescoffier: ah sorry!

neo4j-quickstart/pom.xml
```diff
@@ -12,6 +12,8 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.target>1.8</maven.compiler.target>
+    <docker-plugin.version>0.28.0</docker-plugin.version>
```
cescoffier: Can be removed no?

mcserra: Sure. It's done.

cescoffier:
LGTM, waiting for CI! 

Thanks for this work, much appreciated!

geoand:
Unfortunately this needs to be reverted I think...

Since the merge we have had the native tests cron job fail (see [this](https://github.com/quarkusio/quarkus/issues/6588#issuecomment-594148190)) and a Jenkins instance running the tests report the failure.

I was also able reproduce it locally - so I am opening a PR to revert this. We basically need a green CI this close to a release and can work on fixing the issue at leisure once this is reverted

PR #477, opened 2020-03-02
RSS Regression: Upload pmap output to cloud storage, compare between runs

johnaohara:
The PR aims to make the RSS regression check more robust for the native image cron job by;

1. Capturing pmap output from getting-started native binary
2. Downloading pmap output from previous build from IBM cloud storage
3. Comparing RSS from current build to previous build. If the RSS has grown more than 10%, then the run fails the RSS regression test
4. An upper limit of 40MB is defined to stop unbounded RSS growth by less than 10% each run
5. Current run pmap output is uploaded to IBM cloud storage for comparison against subsequent test

**In order for this PR to function, the following secrets needs to be defined for the repo before this PR is merged**;

- BUCKET_NAME
- API_KEY
- SERVICE_INSTANCE_ID

### DIFF
--- a/.github/RssRegression.java
+++ b/.github/RssRegression.java
@@ -73,7 +73,6 @@ public class RssRegression {
             rssRegression = newRss > (oldRss * 1.1);
             System.out.println("Regression: " + rssRegression);
 
-            System.out.println("Uploading new pmap file to storage: " + bucketFile + " from local: " + oldPmapFile);
         }
         _cosClient.shutdown();
 
@@ -81,6 +80,7 @@ public class RssRegression {
             exit(1);
         }
         else {
+            System.out.println("Uploading new pmap file to storage: " + bucketFile + " from local: " + oldPmapFile);
             uploadObject(_cosClient, BUCKET_NAME, bucketFile, newPmapFile);
             exit(0);
         }
--- a/.github/RssRegression.java
+++ b/.github/RssRegression.java
@@ -1,0 +1,182 @@
+//usr/bin/env jbang "$0" "$@" ; exit $?
+//DEPS com.ibm.cos:ibm-cos-java-sdk:2.1.0
+//DEPS javax.xml.bind:jaxb-api:2.4.0-b180830.0359
+
+package io.quarkus;
+
+import com.ibm.cloud.objectstorage.ClientConfiguration;
+import com.ibm.cloud.objectstorage.SDKGlobalConfiguration;
+import com.ibm.cloud.objectstorage.auth.AWSCredentials;
+import com.ibm.cloud.objectstorage.auth.AWSStaticCredentialsProvider;
+import com.ibm.cloud.objectstorage.client.builder.AwsClientBuilder.EndpointConfiguration;
+import com.ibm.cloud.objectstorage.oauth.BasicIBMOAuthCredentials;
+import com.ibm.cloud.objectstorage.services.s3.AmazonS3;
+import com.ibm.cloud.objectstorage.services.s3.AmazonS3ClientBuilder;
+import com.ibm.cloud.objectstorage.services.s3.model.AmazonS3Exception;
+import com.ibm.cloud.objectstorage.services.s3.model.GetObjectRequest;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+
+import static java.lang.System.exit;
+
+public class RssRegression {
+
+    private static AmazonS3 _cosClient;
+    private static String ENDPOINT_URL;
+    private static String LOCATION;
+
+    static {
+        ENDPOINT_URL = "https://s3.eu-gb.cloud-object-storage.appdomain.cloud";
+        SDKGlobalConfiguration.IAM_ENDPOINT = "https://iam.cloud.ibm.com/identity/token";
+        LOCATION = "eu-gb";
+    }
+
+    /**
+     * @param args
+     */
+    public static void main(String[] args) {
+
+        if(args.length != 6){
+            System.out.println(args.length);
+            printUsage();
+            exit(1);
+        }
+        String BUCKET_NAME = args[0];
+        String API_KEY = args[1];
+        String SERVICE_INSTANCE_ID = args[2];
+        String bucketFile = args[3];
+        String oldPmapFile = args[4];
+        String newPmapFile = args[5];
+
+        _cosClient = createClient(API_KEY, SERVICE_INSTANCE_ID, ENDPOINT_URL, LOCATION);
+
+        System.out.println("Downloading file from storage: " + bucketFile + " to local: " + oldPmapFile);
+        try{
+            downloadFile(_cosClient, BUCKET_NAME, bucketFile, oldPmapFile);
+        } catch (AmazonS3Exception s3Execpetion){
+            System.out.println("Download failed with: " + s3Execpetion.getMessage());
+            _cosClient.shutdown();
+            exit(1);
+        }
+
+        int oldRss = getRssFromFile(oldPmapFile);
+        int newRss = getRssFromFile(newPmapFile);
+        boolean rssRegression = false;
+
+        if( oldRss != -1 && newRss != -1) {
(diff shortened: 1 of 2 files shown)


gsmet:
@johnaohara any chance you could make the secret names a bit more specific so that we know what they are used for?

johnaohara:
@gsmet how about renaming to;

BUCKET_NAME -> IBM_COS_BUCKET_NAME
API_KEY -> IBM_COS_SERVICE_CREDENTIAL_API_KEY
SERVICE_INSTANCE_ID -> IBM_COS_SERVICE_CREDENTIAL_RESOURCE_INSTANCE_ID

I think from those names it should be possible to determine the values required

.github/workflows/native-build-development.yml
```diff
@@ -43,21 +43,40 @@ jobs:
...
+          SERVICE_INSTANCE_ID: ${{ secrets.SERVICE_INSTANCE_ID }}
+
         run: |
-          getting-started/target/getting-started-${QUICKSTART_VERSION}-runner -Xmx2m > server.log &
+          sudo apt-get update -o Dir::Etc::sourcelist="sources.list" -o Dir::Etc::sourceparts="-" -o APT::Get::List-Cleanup="0"
+          sudo apt-get install -y gnupg2 gnupg-agent
+          echo Installing SDKMAN
+          curl -s "https://get.sdkman.io" | bash
+          source ~/.sdkman/bin/sdkman-init.sh
+          sdk install jbang
```
cescoffier: You really want to make Max happy, right?

.github/workflows/native-build-development.yml
```diff
@@ -43,21 +43,40 @@ jobs:
       - name: Check RSS
         env:
           QUICKSTART_VERSION: "1.0-SNAPSHOT"
-          RSS_THRESHOLD: "25000"
+          RSS_THRESHOLD: "40000"
           START_TIME_THRESHOLD: "1000"
+          BUCKET_NAME: ${{ secrets.BUCKET_NAME }}
+          REMOTE_OBJECT_NAME: "pmap.out"
+          API_KEY: ${{ secrets.API_KEY }}
```
cescoffier: We may very soon have multiple API_KEY, can you prefix it with something like S3_API_KEY?

.github/workflows/native-build-development.yml
```diff
@@ -43,21 +43,40 @@ jobs:
       - name: Check RSS
         env:
           QUICKSTART_VERSION: "1.0-SNAPSHOT"
-          RSS_THRESHOLD: "25000"
+          RSS_THRESHOLD: "40000"
           START_TIME_THRESHOLD: "1000"
+          BUCKET_NAME: ${{ secrets.BUCKET_NAME }}
```
cescoffier: I would also prefix this one with S3_

.github/workflows/native-build-development.yml
```diff
@@ -43,21 +43,40 @@ jobs:
...
+          REMOTE_OBJECT_NAME: "pmap.out"
+          API_KEY: ${{ secrets.API_KEY }}
+          SERVICE_INSTANCE_ID: ${{ secrets.SERVICE_INSTANCE_ID }}
```
cescoffier: I believe it's also S3 related, could you prefix it?

geoand:
I personally like them

.github/workflows/native-build-development.yml
```diff
@@ -43,21 +43,40 @@ jobs:
...
           ps --no-headers  -C getting-started-${QUICKSTART_VERSION}-runner -o rss > rss.out
           rss=$(< rss.out)
-          echo "Threshold (kB): ${RSS_THRESHOLD}"
+          echo "Max RSS Threshold (kB): ${RSS_THRESHOLD}"
           echo "RSS (kB): $rss"
           startTime=$(grep -oP "started in \K([0-9]*\.[0-9]*)" server.log)
           startTimeMs=$(echo "$startTime*1000/1" | bc)
           echo "Start time Threshold (ms): ${START_TIME_THRESHOLD}"
           echo "Start time (ms): $startTimeMs"
+          jbang .github/RssRegression.java ${BUCKET_NAME} ${API_KEY} ${SERVICE_INSTANCE_ID} ${REMOTE_OBJECT_NAME} pmap.old.out pmap.out
```
cescoffier: Maybe indicate somewhere that this retrieves the previous pmap and compare it to the new one.

cescoffier: BTW you fails the build it the ration is greater to 1.1 between the old and new one. However, you upload the new one anyway (even if the build failed). So it's unlikely we can catch the regression in the next builds.

.github/workflows/native-build-development.yml
cescoffier: or auth centric maybe.

.github/workflows/native-build-development.yml
johnaohara: Just finding it useful :)

.github/workflows/native-build-development.yml
johnaohara: @gsmet requested changes to the secret names. I proposed;

BUCKET_NAME -> IBM_COS_BUCKET_NAME
API_KEY -> IBM_COS_SERVICE_CREDENTIAL_API_KEY
SERVICE_INSTANCE_ID -> IBM_COS_SERVICE_CREDENTIAL_RESOURCE_INSTANCE_ID

.github/workflows/native-build-development.yml
johnaohara: see previous comment

.github/workflows/native-build-development.yml
johnaohara: see previous comment

.github/workflows/native-build-development.yml
johnaohara: We should probably make that 1.1 configurable. 

What happens in the case that the +10% RSS is genuinely related to an increase as a result of the build tools, i.e. not a regression caused by a bug. It could be valid, all though it requires manual validation, hence the failure and should be compared against the next builds.

johnaohara: Having said that, RSS regression tripping would be the most common failure path, and I am archiving the pmap output with every run. So if it does trip, and is genuinely due to the tooling, we could manually replace the pmap file stored in the cloud bucket

johnaohara: Have changed to upload only if does not regress

.github/RssRegression.java
```diff
@@ -0,0 +1,182 @@
...
+        boolean rssRegression = false;
+
+        if( oldRss != -1 && newRss != -1) {
+            System.out.println("Old RSS: " + oldRss);
+            System.out.println("New RSS: " + newRss);
+
+            rssRegression = newRss > (oldRss * 1.1);
+            System.out.println("Regression: " + rssRegression);
+
+            System.out.println("Uploading new pmap file to storage: " + bucketFile + " from local: " + oldPmapFile);
```
cescoffier: This line should be moved no?

johnaohara: moved that log statement

.github/RssRegression.java
```diff
@@ -0,0 +1,182 @@
...
+        SDKGlobalConfiguration.IAM_ENDPOINT = "https://iam.cloud.ibm.com/identity/token";
+        LOCATION = "eu-gb";
+    }
+
+    /**
+     * @param args
+     */
+    public static void main(String[] args) {
+
+        if(args.length != 6){
```
maxandersen: fyi you can greatly simplify the parsing and usage text by using picocli ... not a blocker for the PR but just a pointer.  example: https://github.com/maxandersen/jbang/blob/master/examples/smee.java#L51

johnaohara: Thanks for the pointer

johnaohara:
secretes have been added to repo, merging

.github/workflows/native-build-development.yml
maxandersen: fyi, you could have used the maxandersen/jbang-action docker image here

PR #483, opened 2020-03-05
Make GreetingResource independent of Resteasy

gasgithub:
Absolutely no reason to depend on reasteasy in this class.

**Check list**:

Your pull request:

- [ ] makes sure the documentation must not be updated


cescoffier:
We are using the RestEasy one on purpose because it does not require passing the "name" of the parameter explicitly.

gasgithub:
I can understand that, but from my point of view it has too little value to make the code framework dependent just to save 8 chars of code.

gsmet:
@gasgithub you can use what you want in your own app. But we try to prefer brevity in the Quarkus quickstarts. Thanks anyway.

PR #495, opened 2020-03-13
Fixing and reordering imports

rsvoboda:
Fixing and reordering imports

Result of `mvn -fae -Dproject.build.sourceEncoding=UTF-8 net.revelc.code:impsort-maven-plugin:1.3.2:sort -Dimpsort.groups='java.,javax.,org.,com.' -Dimpsort.staticGroups='*' -Dimpsort.removeUnused=true`

### DIFF
--- a/hibernate-search-elasticsearch-quickstart/src/test/java/org/acme/hibernate/search/elasticsearch/service/LibraryResourceInGraalIT.java
+++ b/hibernate-search-elasticsearch-quickstart/src/test/java/org/acme/hibernate/search/elasticsearch/service/LibraryResourceInGraalIT.java
@@ -15,20 +15,6 @@
  */
 package org.acme.hibernate.search.elasticsearch.service;
 
-/*
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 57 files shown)


hibernate-search-elasticsearch-quickstart/src/test/java/org/acme/hibernate/search/elasticsearch/service/LibraryResourceInGraalIT.java
```diff
@@ -15,6 +15,21 @@
  */
 package org.acme.hibernate.search.elasticsearch.service;
 
+/*
```
cescoffier: That looks like too many license headers...

rsvoboda: yeah, nice catch :)

rsvoboda: PR updated

PR #496, opened 2020-03-13
parameterize two missed quickstarts

maxandersen:
make the two new quickstarts parameterized.


cescoffier:
ah ah... told you we got now ones :-D

LGTM, waiting for CI.

PR #505, opened 2020-03-19
OptaPlanner quickstart: demo UI improvements

ge0ffrey:
This doesn't impact the quarkus backend.
Just makes the demo UI a bit more readable, as well as clean up some of the JS code, to improve the demo experience.


ge0ffrey:
Hmm, this seems to be directed to "master" instead of "development". Closing this PR and resubmitting correctly. (not that it would harm master, this is not the right way and there is no need to backport this)

PR #526, opened 2020-04-16
Introduce docker pruning when running the cron job

geoand:


docker-prune.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# pruning is only run when inside CI to avoid accidentally removing stuff
+# GITHUB_ACTIONS is always set to true inside Github Actions
+# https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables
+if [ "${GITHUB_ACTIONS}" == true ] ; then
+  docker container prune -f
+  docker image prune -f
+  docker network prune -f
+  docker volume prune -f
+fi
```
gsmet, line 13: Same here, maybe move it to `.github`?

cescoffier: Yes, it would be better to not show that to innocent eyes ;-)

geoand:
PR Updated with the proper location

PR #527, opened 2020-04-17
Anonymous user is now represented by Principal with an empty name

sberyozkin:
Principal with an empty name is probably better than a null Principal (and it is probably more common). https://github.com/quarkusio/quarkus/pull/8556 itself does not change the way the name is handled.
I think we may need to mention it in the migration guide


gsmet:
Hmmm, this is a very important change potentially breaking applications in a weird way.

@stuartwdouglas is it a conscious one?

sberyozkin:
It is actually changed at https://github.com/quarkusio/quarkus/pull/8556/files#diff-3217632482f5fdf813a5f550051571b6.
As far as I recall this is more correct than returning a null principal, but indeed, there could be users who check a null principal, so if this change stays then the migration guide update will be needed.

stuartwdouglas:
This is a bug, it needs to be fixed.

stuartwdouglas:
https://github.com/quarkusio/quarkus/pull/8678 is the fix

Issue #521, opened 2020-04-11
Update documentation and example tp show how to secure rest endpoint with oauth2 token scopes.


mishraomp:
There is no documentation around, how to secure a rest resource based on jwt token scope using annotation, the only annotation we can use is ` @RolesAllowed({"WRITE_DATA"})`. 

would like to know how to secure with scope.
{
  "jti": "1fbc98bf-16dd-4b19-b6a6-725fc9949d96",
  "exp": 1586572797,
  "nbf": 0,
  "iat": 1586570997,
  "iss": "**********************",
  "aud": "account",
  "sub": "82558a4b-40f9-416f-8475-d25cc51806b6",
  "typ": "Bearer",
  "azp": "soam-api-service",
  "auth_time": 0,
  "session_state": "11f93c8e-ef77-46f2-80e8-8539c24f21ce",
  "acr": "1",
  "realm_access": {
    "roles": [
      "offline_access",
      "uma_authorization"
    ]
  },
  "resource_access": {
    "account": {
      "roles": [
        "manage-account",
        "manage-account-links",
        "view-profile"
      ]
    }
  },
  "scope": "profile WRITE_DATA",
  "email_verified": false
}

PR #529, opened 2020-04-17
Add openid connect quickstart test

### LINKED ISSUE #355: Add tests to the openid connect quickstart
* Add Quarkus tests
* Add native tests

cemnura:
Fixes #355

**Check list**:

Your pull request:

- [x] has native tests (`mvn clean verify -Pnative`)

I would also enjoy improving the documentation for oauth2 mentioned in  #521

### DIFF
--- a/security-openid-connect-quickstart/pom.xml
+++ b/security-openid-connect-quickstart/pom.xml
@@ -14,6 +14,7 @@
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
         <quarkus.platform.version>999-SNAPSHOT</quarkus.platform.version>
+        <keycloak.version>11.0.1</keycloak.version>
         <surefire.version>3.0.0-M5</surefire.version>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
@@ -41,16 +42,21 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy-jsonb</artifactId>
         </dependency>
-
-        <!-- Test -->
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>io.rest-assured</groupId>
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-junit5</artifactId>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.12.3</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -64,6 +70,7 @@
                     <systemPropertyVariables>
                         <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
                         <maven.home>${maven.home}</maven.home>
+                        <keycloak.version>${keycloak.version}</keycloak.version>
                     </systemPropertyVariables>
                 </configuration>
             </plugin>
@@ -125,7 +132,36 @@
                                             org.jboss.logmanager.LogManager
                                         </java.util.logging.manager>
                                         <maven.home>${maven.home}</maven.home>
+                                        <keycloak.version>${keycloak.version}</keycloak.version>
                                     </systemPropertyVariables>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <id>docker</id>
+            <activation>
+                <property>
+                    <name>docker</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>docker-prune</id>
+                                <phase>generate-resources</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <executable>${basedir}/../.github/docker-prune.sh</executable>
                                 </configuration>
                             </execution>
                         </executions>
--- a/security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/BearerTokenAuthenticationTest.java
+++ b/security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/BearerTokenAuthenticationTest.java
@@ -1,0 +1,52 @@
+package org.acme.security.openid.connect;
+
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@QuarkusTestResource(KeycloakServer.class)
+public class BearerTokenAuthenticationTest {
+
+    @Test
+    public void testAdminAccess() {
+        RestAssured.given().auth().oauth2(getAccessToken("admin"))
+                .when().get("/api/admin")
+                .then()
+                .statusCode(200);
+
+        RestAssured.given().auth().oauth2(getAccessToken("alice"))
+                .when().get("/api/admin")
+                .then()
+                .statusCode(403);
+    }
+
+    @Test
+    public void testUserAccess() {
+
+        RestAssured.given().auth().oauth2(getAccessToken("alice"))
+                .when().get("/api/users/me")
+                .then()
+                .statusCode(200);
+
+        RestAssured.given().auth().oauth2(getAccessToken("admin"))
+                .when().get("/api/users/me")
+                .then()
+                .statusCode(200);
+    }
+
+    private String getAccessToken(String userName) {
+        return RestAssured
+                .given()
+                .param("grant_type", "password")
+                .param("username", userName)
+                .param("password", userName)
+                .param("client_id", "backend-service")
+                .param("client_secret", "secret")
+                .when()
+                .post("http://localhost:8180/auth/realms/quarkus/protocol/openid-connect/token")
+                .jsonPath().get("access_token");
+    }
+}
--- a/security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/KeycloakServer.java
+++ b/security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/KeycloakServer.java
@@ -1,0 +1,35 @@
+package org.acme.security.openid.connect;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import org.testcontainers.containers.BindMode;
+import org.testcontainers.containers.FixedHostPortGenericContainer;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class KeycloakServer implements QuarkusTestResourceLifecycleManager {
+
+    private GenericContainer keycloak;
+
+    @Override
+    public Map<String, String> start() {
+        keycloak = new FixedHostPortGenericContainer("quay.io/keycloak/keycloak:" + System.getProperty("keycloak.version"))
+                .withFixedExposedPort(8180, 8080)
+                .withEnv("DB_VENDOR", "H2")
+                .withEnv("KEYCLOAK_USER", "admin")
+                .withEnv("KEYCLOAK_PASSWORD", "admin")
+                .withEnv("KEYCLOAK_IMPORT", "/tmp/realm.json")
+                .withClasspathResourceMapping("quarkus-realm.json", "/tmp/realm.json", BindMode.READ_ONLY)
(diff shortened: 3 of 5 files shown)


security-oauth2-quickstart/src/main/resources/application.properties
```diff
@@ -1,3 +1,4 @@
-quarkus.oauth2.client-id=client_id
+quarkus.oauth2.enabled=true
+quarkus.oauth2.client-id=backend-service
 quarkus.oauth2.client-secret=secret
-quarkus.oauth2.introspection-url=http://oauth-server/introspect
\ No newline at end of file
+quarkus.oauth2.introspection-url=http://localhost:8180/auth/realms/quarkus/protocol/openid-connect/token/introspect
```
cescoffier, line 4: Wondering if this property can use a random port (and set by a Quarkus Test Resource)

security-oauth2-quickstart/src/test/java/org/acme/security/oauth2/TokenSecuredResourceTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+@QuarkusTest
+public class TokenSecuredResourceTest {
+
+    private static final String KEYCLOAK_SERVER_URL = System.getProperty("keycloak.url", "http://localhost:8180/auth");
```
cescoffier, line 18: So we cannot use TestContainer like this, as the lifecycle is different from the Quarkus lifecycle.

You need to use a Quarkus Test Resource, and BTW, it will also let you set some properties dynamically. 
Check how the Kafka quickstart is built as an example.

security-oauth2-quickstart/src/test/java/org/acme/security/oauth2/TokenSecuredResourceTest.java
```diff
@@ -0,0 +1,76 @@
...
+    private static final String KEYCLOAK_REALM = "quarkus";
+
+    @ClassRule
+    @Container
+    private static GenericContainer keycloak = new FixedHostPortGenericContainer("quay.io/keycloak/keycloak:9.0.3")
```
cescoffier, line 23: This should be moved to a Quarkus Test Resources, and probably bound to a free port. 

Also, I'm surprised about @ClassRule as it's a Junit 4 annotation.

security-oauth2-quickstart/src/test/resources/.notes
```diff
@@ -0,0 +1,66 @@
+curl -X POST \
```
cescoffier, line 1: What the goal of this file?

cescoffier (dismissed):
We cannot use the  `@TestContainer` as it introduces a lifecycle mismatch. You need to define a Quarkus Test Resource (and use the TestContainer API in it).

cescoffier:
@pedroigor @sberyozkin  can you review the "security" part? There is an endpoint to retrieve tokens. While I understand the usage in tests, it should be mentioned somewhere to not use this in production (right?)

security-oauth2-quickstart/src/test/resources/.notes
cemnura: These where my notes when developing this qucikstart. I thought that I may create documentation using these commands.

security-oauth2-quickstart/src/test/java/org/acme/security/oauth2/TokenSecuredResourceTest.java
cemnura: Okey I will move try to move this and try to bound to a free port.

Yes you are correct ``@ClassRule`` is a Junit4 annotation. I misread the testcontainer documentation. I will be removing this in general.

security-oauth2-quickstart/src/test/java/org/acme/security/oauth2/TokenSecuredResourceTest.java
cemnura: I will check how the Kafka quickstart is built.

security-oauth2-quickstart/src/main/resources/application.properties
cemnura: This configuration is overwritten during at QuarkusTestResource creation. I was not able to delete this configuration because the following error occurs if not written.
```
io.quarkus.runtime.configuration.ConfigurationException: client-id, client-secret and introspection-url must be configured when the oauth2 extension is enabled
```

security-oauth2-quickstart/src/test/resources/.notes
cemnura: I will delete these files when I am going to squash commits.

cemnura:
> We cannot use the `@TestContainer` as it introduces a lifecycle mismatch. You need to define a Quarkus Test Resource (and use the TestContainer API in it).

I have defined a Quarkus Test Resource to enable the creation of a keycloak server for testing.

sberyozkin:
@cescoffier the `oauth2` extension was added by Loic originally to support the opaque (binary) tokens, they should be available to the service code as `SecurityIdentity.getCredential(TokenCredential.class)` - it is probably up to the application what to do with this token if they want to propagate.
@cemnura `quarkus-oidc`, `service` application type - JWT will be verified by the locally cached JWK keys while binary tokens should be introspected (though we don't have a test for the binary token case)

cemnura:
Hello @sberyozkin, I did not completely understand you. Do you think there are missing test?

loicmathieu:
@cemnura just saw your PR after I created #609 to add a test and a mock of the authentication server via WireMock.

Reading your PR, it appears that this is an OIDC based test, it should be better to add it to the `security-openid-connect-quickstart` quickstart that also miss some test as the OAuth2 extension is a lightweight security implementation for OAuth2 opaque token (no JWT, no OIDC).

pedroigor:
@loicmathieu I agree. It seems that the tests herein are basically checking whether or not the extension is capable of also introspecting JWT tokens if comparing to yours. Although still a valid OAuth2 usage (you may have authorization servers issuing JWTs too). Your tests seem to be simple as they don't require a Keycloak container running. One thing I'm missing though is checks for failures when the introspection endpoint says that the token is not active.

And yeah, a similar test to oidc quickstart is necessary.

cemnura:
Hello @loicmathieu thanks for viewing this PR. Would you think this unit test would be better in a different quickstart?

@pedroigor I aimed to test the quickstart by testing oauth2 issued tokens from keycloak and send them to the rest api via "Authorization: Bearer" header to validate RBAC. 

The roles for the users are in the  quarkus-realm.json which is a keycloak realm import file.

loicmathieu:
@cemnura yes, for me it's a valid integration test for the OIDC quickstart, so moving your code to `security-openid-connect-quickstart` seems a good idea.

@sberyozkin is the guardian of OIDC stuff, so better check with him ;)
I think he said that the same Wiremock setup could be used also for OIDC instead of using a real keycloack server. 
So maybe this would be abandonned, sorry for that ...

Anyway, thanks for trying to improve our quickstart ;)

cemnura:
@loicmathieu no problem at all the I believe the Wiremock approach would be a lighter test then starting a keycloak server with docker also making the test dependent to docker.

I will checkout Wiremock for myself as well thanks for the information 🙂 

@sberyozkin I would happily move/improve this integration test to `security-openid-connect-quickstart` as @loicmathieu stated if you see fit.

cemnura:
@sberyozkin I have converted the test to the `security-openid-connect-quickstart` quickstart project using TestContainer approach.

I added a Keycloak mock server using WireMock for evaluation and discussion. This still needs work since the server still does not respond to the `protocol/openid-connect/token`

sberyozkin:
Hi @cemnura Thanks, it looks good, but I don't think the wiremock code (as well as the docker profile) should be in this PR at all. Please remove it.
Can I suggest, the following, copy and paste `integration-tests/oidc` to `integration-tests/oidc-wiremock` in the main Quarkus respository (and update `integration-tests/pom.xml`) and in this new more module, in its pom, remove the docker and keycloak profiles, remove `KeycloakTestManager` (which sets up keycloak realms/clients) and have this test working with WireMock only

cemnura (commented):
> Hi @cemnura Thanks, it looks good, but I don't think the wiremock code (as well as the docker profile) should be in this PR at all. Please remove it.
> Can I suggest, the following, copy and paste `integration-tests/oidc` to `integration-tests/oidc-wiremock` in the main Quarkus respository (and update `integration-tests/pom.xml`) and in this new more module, in its pom, remove the docker and keycloak profiles, remove `KeycloakTestManager` (which sets up keycloak realms/clients without Wiremoce) and have this test working with WireMock only (as you have shown here)



> Hi @cemnura Thanks, it looks good, but I don't think the wiremock code (as well as the docker profile) should be in this PR at all. Please remove it.
> Can I suggest, the following, copy and paste `integration-tests/oidc` to `integration-tests/oidc-wiremock` in the main Quarkus respository (and update `integration-tests/pom.xml`) and in this new more module, in its pom, remove the docker and keycloak profiles, remove `KeycloakTestManager` (which sets up keycloak realms/clients without Wiremoce) and have this test working with WireMock only (as you have shown here)

Okey I am updating the PR by removing the WireMock code then migrating it to the new `integration-tests/oidc-wiremock`
as you have recommended.

Will reference this PR in the main Quarkus repository fir refrerence.

security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/KeycloakTestResource.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Map;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.options;
+
+public class KeycloakTestResource implements QuarkusTestResourceLifecycleManager {
```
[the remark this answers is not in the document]
cemnura: I have deleted this class from this PR and moving it to the main quarkus repository

sberyozkin:
@cemnura Can you please rebase and force-push again, as I've just removed `WIP:` in the PR name.

security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/UsersResourceTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@QuarkusTestResource(KeycloakServer.class)
+public class UsersResourceTest {
```
sberyozkin, line 11: @cemnura Can you please have a single test class (say `BearerTokenAuthenticationTest`) combining both admin and user resources tests - to have KC started and stopped once and avoid some instability.

cemnura: @sberyozkin gladly. Resolving ASAP.

cemnura: I combined the test classes into one class being `BearerTokenAuthenticationTest`.

cemnura:
I added a `@NativeImageTest` to satisfy PR checklist.

- [ ] has native tests (mvn clean verify -Pnative)

However, the native image test is failing. I have found the reason to be;

```java

    @ConfigProperty(name = "quarkus.oidc.auth-server-url")
    String oidc;
```
 
in the test class is set as `null` string during `mvn clean verify -Pnative` leading to the following error when trying to get access token;

```
  <body>RESTEASY003210: Could not find resource for full path: http://localhost:8081/null/protocol/openid-connect/token</body>
```

What would your recommendation be?

sberyozkin:
@cemnura Can you please use this version of `KeycloakServer`:
https://github.com/quarkusio/quarkus-quickstarts/blob/master/security-openid-connect-web-authentication-quickstart/src/test/java/org/acme/security/openid/connect/web/authentication/KeycloakServer.java
(Note it uses a fixed port), and then in the test use a keycloak realm URL with `8180`. (This is also done in `security-openid-connect-web-authentication-quickstart/.../application.properties`)

The way you inject it is interesting, can you please open a main Quarkus repository issue (something like `QuarkusTestResourceLifecycleManager properties can not be injected in the native image test`) and CC it to @stuartwdouglas ? This would be a nice feature...

Thanks

cemnura:
@sberyozkin I originally implemented a fixed port and reorganized TestContainer to use a random port according to previous recommendations. 
I would gladly reuse the;
https://github.com/quarkusio/quarkus-quickstarts/blob/master/security-openid-connect-web-authentication-quickstart/src/test/java/org/acme/security/openid/connect/web/authentication/KeycloakServer.java

I already tried it with native test and it definitely works.

 I will update the PR today.

> The way you inject it is interesting, can you please open a main Quarkus repository issue (something like QuarkusTestResourceLifecycleManager properties can not be injected in the native image test) and CC it to @stuartwdouglas ? This would be a nice feature...

Honestly I had no idea that it would even work I just tired it on a hence and it surprisingly worked. But it does not work with `@NativeImageTest`. I will create an issue regarding the usage of `@ConfigProperties` to be injected during `@QuarkusTest`

cemnura:
I removed the `@ConfigProperty` field in the test and made the TestContainer port fixed and made the test pass in both JVM and Native mode.

sberyozkin (approved):
Nice work @cemnura, thanks

PR #539, opened 2020-04-24
Add application.properties for easy configuration

maxandersen:
Why:

 * getting-started is used by default in Eclipse Che and
   currently user need to know where to put the application.properties
   to start configuring. Would be nice if file is already in place.

This change addreses the need by:

 * Adds empty/default `application.properties`

Related to https://github.com/eclipse/che/issues/16700

### DIFF
--- a/getting-started/src/main/resources/application.properties
+++ b/getting-started/src/main/resources/application.properties
@@ -1,0 +1,2 @@
+# Quarkus Configuration file
+# key = value


maxandersen:
rebased against development branch. Anything else prevent this from being applied ?

Sanne:
merged, thanks @maxandersen !

PR #540, opened 2020-04-24
initial command mode quickstart

maxandersen:
Adds command mode quickstart.

No tests and .s2i as couldn't find what would be reasonable to add there :)


getting-started-command-mode/pom.xml
```diff
@@ -0,0 +1,123 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project>
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.acme</groupId>
+    <artifactId>getting-started-command-mode</artifactId>
+    <version>1.0-SNAPSHOT</version>
+
+    <properties>
+        <quarkus-plugin.version>999-SNAPSHOT</quarkus-plugin.version>
```
sunix, line 10: it has to be 1.4.0.Final or 1.4.1.Final, right? not 999-SNAPSHOT

maxandersen: not in development branch.

See checklist when opening a pr:

if you are asking in general - then yes, this feature was first released in 1.4.0.Final.

maxandersen:
@cescoffier can we get these in to update docs do release blog with ref to this repo ? 

the build errors are totally unrelated to this PR afaics.

getting-started-command-mode/src/main/docker/Dockerfile.native
```diff
@@ -0,0 +1,22 @@
+####
```
cescoffier, line 1: Do we really need dockerfiles for command line?

getting-started-command-mode/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+# Quarkus Configuration file
```
cescoffier, line 1: Can be removed.

maxandersen: this is what is there in other quickstarts and makes it so users can just open the file and tweak without having to remember where it should be put....so I kept it. Objection?

getting-started-command-mode/src/main/docker/Dockerfile.native
maxandersen: if you want to build a native binary for your linux container when running on windows or osx then you do.

maxandersen:
> I would prefer to have at least one test if possible. But I'm not sure we can?

I can't really think of any good way of doing it without needing something else than junit

getting-started-command-mode/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+        <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+        <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+        <quarkus.platform.version>1.4.0.Final</quarkus.platform.version>
```
sunix, line 13: So should be 999-SNAPSHOT here?

maxandersen: yes, good catch. fixed!

getting-started-command-mode/src/main/docker/Dockerfile.native
cescoffier: right! Make sense... How to work around the lack of cross-compilation! Docker!

getting-started-command-mode/src/main/resources/application.properties
cescoffier: No, it's fine. Yes, the other quickstarts still have them.

cescoffier:
All good for me, merging!

PR #549, opened 2020-04-30
Use disk space reclaim script from quarkus repo

geoand:
Same as https://github.com/quarkusio/quarkus/pull/8963


cescoffier:
This would need to be backported to the development branch, right?

geoand:
It's not strictly necessary since the cron job uses the definition from `master`, but it would nonetheless be good to do so.

geoand:
So if you could also cherry-pick the commit into development, that would be great (I don't seem to have the karma to do so).

gsmet:
The master branch is rewritten regularly so yes we need it in the development branch too.

Will take care of it.

geoand:
Thanks!

PR #404, opened 2019-12-13
Extension guide

danielpetisme:
**Check list**:

Your pull request:

- [x] targets the `development` branch
- [x] uses the `999-SNAPSHOT` version of Quarkus
- [x] has tests (`mvn clean test`)
- [ ] works in native (`mvn clean package -Pnative`)
- [ ] has native tests (`mvn clean verify -Pnative`)
- [ ] makes sure the documentation must not be updated
- [ ] links the documentation update pull request (if needed)
- [ ] updates or creates the `README.md` file (with build and run instructions)
- [ ] For new quickstart, is located in the directory _component-quickstart_
- [ ] For new quickstart, is added to the root `pom.xml` and `README.md`

PR #584, opened 2020-05-26
Add getting-started-extension quickstart

gsmet:
I revisited https://github.com/quarkusio/quarkus-quickstarts/pull/404 to enforce the proper versions via the bom as everything was incorrect and using random versions.

Not sure about the structure tbh but since we have the guides, I think we should merge it if it passes the tests.


cescoffier:
@gsmet How should we proceed? It would be a nice addition.

PR #605, opened 2020-06-18
Apply minor fixes to command-mode quickstart

geoand:


gsmet:
Both CI failed, could you check?

geoand:
The CI failures were on a different module - I just rebases so let's see what happens

PR #612, opened 2020-07-02
Use Mutiny event-based API in the Hibernate Reactive quickstart

DavideD:
**WARNING**: It requires this PR in Quarkus: https://github.com/quarkusio/quarkus/pull/10433

Update the Hibernate Reactive quickstart so that it uses the Mutiny event-based API.
It also inject the session directly making the code simpler.

@Sanne, @gavinking Would you like to have a look?

### DIFF
--- a/hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
+++ b/hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
@@ -1,7 +1,11 @@
 package org.acme.hibernate.reactive;
 
+import java.util.List;
+import java.util.function.Function;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.json.Json;
+import javax.json.JsonObjectBuilder;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
@@ -11,11 +15,15 @@ import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
 
 import org.hibernate.reactive.mutiny.Mutiny;
 
+import org.jboss.logging.Logger;
 import org.jboss.resteasy.annotations.jaxrs.PathParam;
 
+import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
 
 @Path("fruits")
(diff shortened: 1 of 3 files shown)


hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
```diff
@@ -25,65 +25,67 @@
...
-    Uni<Mutiny.Session> mutinySession;
+    Mutiny.Session mutinySession;
 
     @GET
     public Uni<Fruit[]> get() {
-        return mutinySession.flatMap( session -> session.createNamedQuery( "Fruits.findAll", Fruit.class ).getResultList() )
-                .map( fruits -> fruits.toArray( new Fruit[fruits.size()] ) );
+        return mutinySession
+                .createNamedQuery( "Fruits.findAll", Fruit.class ).getResultList()
+                .onItem().apply( fruits -> fruits.toArray( new Fruit[fruits.size()] ) );
```
Sanne, line 34: Could the "get" method return `Uni<List<Fruit>>` rather than needing the `toArray` conversion?

hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
```diff
@@ -25,65 +25,67 @@
...
     }
 
     @GET
     @Path("{id}")
     public Uni<Fruit> getSingle(@PathParam Integer id) {
-        return mutinySession
-                .flatMap( session -> session.find( Fruit.class, id ) );
+        return mutinySession.find( Fruit.class, id );
```
Sanne, line 40: great, much better!

hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
```diff
@@ -25,65 +25,67 @@
...
+                .find( Fruit.class, id )
+                // Found
+                .onItem().ifNotNull()
+                    .produceUni( entity -> {
+                        entity.setName( fruit.getName() );
+                        return mutinySession.flush()
+                                .onItem().apply( ignore -> Response.ok( entity ).build() );
+                    } )
+                // Not found
+                .onItem().ifNull()
```
Sanne, line 72: I have to admit that even with help from indentation, it's not clear to me to which element of the chain the "ifNull" applies here..

hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
DavideD: I think I tried and didn't work, I'll have a second look.

hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
DavideD: What do you think it apply to?

DavideD: > What do you think it apply to?

I just realized that I might sound more confrontational than I intended. I'm just trying to understand what you find confusing so that I can think of a better approach

Sanne: Mine was  just a thought about the `ifNull` API, I'm not sure if there is any thing you can do about it.

In particular, it loolks the "if null" check could apply on the last produced item: the `Response` object

DavideD: Do you think something like this is less confusing?

        // Update the entity
        Function<Fruit, Uni<Response>> update = entity -> {
            entity.setName(fruit.getName());
            return mutinySession.flush()
                    .onItem().apply(ignore -> Response.ok( entity ).build());
        };

        return mutinySession
                .find( Fruit.class, id )
                      // If entity exists then
                    .onItem().ifNotNull()
                        .produceUni(update)
                    // else
                    .onItem().ifNull()
                        .continueWith(Response.ok().status( 404 ).build());

DavideD: I guess it's not correct though, this works because update never returns null so the comment is a bit misleading

DavideD: I checked the documentation and this is the way it is supposed to be used so I think we are good: https://smallrye.io/smallrye-mutiny/#_how_do_i_handle_null

I will extract the function because I think it makes the code more readable.

hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
DavideD: Yep, it works with List. Nice.
I think we should update the ORM test as well: https://github.com/quarkusio/quarkus-quickstarts/blob/development/hibernate-orm-quickstart/src/main/java/org/acme/hibernate/orm/FruitResource.java#L37

DavideD:
I pushed force some changes:

* get returns `Uni<List<Fruit>>` instead of `Uni<Fruit[]>` 
* Refactoring the methods using IfNotNull, I think it's more readable now
* Added the exception mapping
* Additional test cases 

I think we should also update the ORM quickstart and return a List instead of an array when we do a find: https://github.com/quarkusio/quarkus-quickstarts/blob/development/hibernate-orm-quickstart/src/main/java/org/acme/hibernate/orm/FruitResource.java#L37

I can do it but do you want me to create a new issue or I can do it in this one?

Thanls

Sanne:
very nice @DavideD ! Let's wait for @gsmet 's green light.

Sanne:
> I think we should also update the ORM quickstart and return a List instead of an array when we do a find: https://github.com/quarkusio/quarkus-quickstarts/blob/development/hibernate-orm-quickstart/src/main/java/org/acme/hibernate/orm/FruitResource.java#L37
> 
> I can do it but do you want me to create a new issue or I can do it in this one?

No need to create an issue, but send a separate PR?

DavideD:
https://github.com/quarkusio/quarkus-quickstarts/pull/613

hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
```diff
@@ -23,67 +30,118 @@
...
 
     @Inject
-    Uni<Mutiny.Session> mutinySession;
+    Mutiny.Session mutinySession;
 
     @GET
-    public Uni<Fruit[]> get() {
-        return mutinySession.flatMap( session -> session.createNamedQuery( "Fruits.findAll", Fruit.class ).getResultList() )
-                .map( fruits -> fruits.toArray( new Fruit[fruits.size()] ) );
+    public Uni<List<Fruit>> get() {
```
burmanm, line 39: This is perhaps the wrong place, but given it's a quickstart to help people.. why isn't this ``Multi<Fruit>`` ? Seems like a weird API if it's only a Uni wrapper for normal blocking models instead of real reactive one.

DavideD: At the moment, hibernate reactive returns a Uni for reasons described here: https://github.com/hibernate/hibernate-reactive/issues/49

But you are right, we could return a Multi in this case. I think I will update the quickstart colling `Session#getResults` instead

DavideD:
Updated the return value of the `getAll` and rebased

gsmet:
@DavideD if this is ready it can be merged. Do you confirm?

DavideD:
I confirm

gsmet:
Merged, thanks!

PR #624, opened 2020-07-21
Bump GraalVM version used in cron job to 20.1

geoand:
Done because the cron job has started failing after we removed support for GraalVM 19.3 (see https://github.com/quarkusio/quarkus/issues/6588)

### DIFF
--- a/.github/workflows/native-build-development.yml
+++ b/.github/workflows/native-build-development.yml
@@ -25,10 +25,10 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
-      - name: Install JDK 8
-        uses: joschi/setup-jdk@v1.0.0
+      - name: Install JDK 11
+        uses: joschi/setup-jdk@v2
         with:
-          java-version: 'openjdk8'
+          java-version: '11'
 
       - name: Build Quarkus master
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/native-build-development.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         run: |
           mvn -B clean install --fail-at-end -Pnative -Ddocker \
             -Dquarkus.native.container-build=true \
-            -Dquarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-native-image:19.3.1-java8 \
+            -Dquarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-native-image:20.1.0-java8 \
```
gsmet, line 43: Shouldn't we switch to the java11 flavor for this and the installed JDK? We made it the default.

geoand: I initially thought that the point of this cron was to run the quickstarts with Java 8, but now I see that that is not the case.

.github/workflows/native-build-development.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         run: |
           mvn -B clean install --fail-at-end -Pnative -Ddocker \
             -Dquarkus.native.container-build=true \
-            -Dquarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-native-image:19.3.1-java8 \
+            -Dquarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-native-image:20.1.0-java11 \
```
gsmet, line 43: Should you switch to JDK 11 here too: https://github.com/quarkusio/quarkus-quickstarts/pull/624/files#diff-6fc556614702dfd6015c4f7b457f47c8R28 ?

geoand: You are right, done

geoand:
:tada:

geoand:
Also, this needs to be cherry-picked into `master`, because that is where the cron runs from

gsmet:
Done!

geoand:
Thanks!

PR #626, opened 2020-07-21
Hibernate reactive routes quickstart improvements

mkouba:
https://github.com/quarkusio/quarkus/pull/10871 must be merged first...

### DIFF
--- a/hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
+++ b/hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
@@ -6,6 +6,7 @@ import static io.vertx.core.http.HttpMethod.GET;
 import static io.vertx.core.http.HttpMethod.POST;
 import static io.vertx.core.http.HttpMethod.PUT;
 
+import java.util.List;
 import java.util.NoSuchElementException;
 
 import javax.inject.Inject;
@@ -15,10 +16,8 @@ import org.jboss.logging.Logger;
 
 import io.quarkus.vertx.web.Body;
 import io.quarkus.vertx.web.Param;
-import io.quarkus.vertx.web.ReactiveRoutes;
 import io.quarkus.vertx.web.Route;
 import io.quarkus.vertx.web.RouteBase;
-import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
 import io.vertx.core.http.HttpServerResponse;
 import io.vertx.core.json.JsonObject;
@@ -36,8 +35,10 @@ public class FruitsRoutes {
     Mutiny.Session session;
 
     @Route(methods = GET, path = "/")
-    public Multi<Fruit> getAll() throws Exception {
-        return ReactiveRoutes.asJsonArray(session.createNamedQuery(Fruit.FIND_ALL, Fruit.class).getResults());
+    public Uni<List<Fruit>> getAll() throws Exception {
+        // In this case, it makes sense to return a Uni<List<Fruit>> because we return a reasonable amount of results
+        // Consider returning a Multi<Fruit> for result streams 
+        return session.createNamedQuery(Fruit.FIND_ALL, Fruit.class).getResultList();
     }
 
     @Route(methods = GET, path = "/:id")
--- a/hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
+++ b/hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
@@ -1,123 +1,112 @@
 package org.acme.hibernate.reactive;
 
-import javax.enterprise.context.ApplicationScoped;
-import javax.inject.Inject;
-
-import org.hibernate.reactive.mutiny.Mutiny;
-
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.quarkus.vertx.web.Route;
-import io.quarkus.vertx.web.RouteBase;
-import io.smallrye.mutiny.Multi;
-import io.smallrye.mutiny.Uni;
-import io.vertx.core.http.HttpServerResponse;
-import io.vertx.ext.web.RoutingContext;
-
+import static io.quarkus.vertx.web.Route.HandlerType.FAILURE;
 import static io.vertx.core.http.HttpMethod.DELETE;
 import static io.vertx.core.http.HttpMethod.GET;
 import static io.vertx.core.http.HttpMethod.POST;
 import static io.vertx.core.http.HttpMethod.PUT;
 
+import java.util.NoSuchElementException;
+
+import javax.inject.Inject;
+
+import org.hibernate.reactive.mutiny.Mutiny;
+import org.jboss.logging.Logger;
+
+import io.quarkus.vertx.web.Body;
+import io.quarkus.vertx.web.Param;
+import io.quarkus.vertx.web.ReactiveRoutes;
+import io.quarkus.vertx.web.Route;
+import io.quarkus.vertx.web.RouteBase;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.Uni;
+import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.RoutingContext;
+
 /**
  * An example using Vert.x Web, Hibernate Reactive and Mutiny.
  */
-@ApplicationScoped
 @RouteBase(path = "/fruits", produces = "application/json")
 public class FruitsRoutes {
 
-    @Inject
-    ObjectMapper mapper;
+    private static final Logger LOGGER = Logger.getLogger(FruitsRoutes.class.getName());
 
     @Inject
     Mutiny.Session session;
 
-    /**
-     * Convert an object into a JSON string.
-     */
-    private String toJSON(Object object) {
-        try {
-            return mapper.writeValueAsString(object);
-        } catch (JsonProcessingException e) {
-            throw new IllegalArgumentException(e);
-        }
-    }
-
     @Route(methods = GET, path = "/")
-    public Multi<Fruit> getAll(RoutingContext rc) throws Exception {
-        return session.createNamedQuery(Fruit.FIND_ALL, Fruit.class).getResults();
+    public Multi<Fruit> getAll() throws Exception {
+        return ReactiveRoutes.asJsonArray(session.createNamedQuery(Fruit.FIND_ALL, Fruit.class).getResults());
     }
 
     @Route(methods = GET, path = "/:id")
-    public Uni<Fruit> getSingle(RoutingContext rc) {
-        final Integer id = Integer.valueOf(rc.request().getParam("id"));
-
-        return session.find(Fruit.class, id);
+    public Uni<Fruit> getSingle(@Param String id) {
+        return session.find(Fruit.class, Integer.valueOf(id));
     }
 
     @Route(methods = POST, path = "/")
-    public Uni<HttpServerResponse> create(RoutingContext rc) {
-        final String name = rc.getBodyAsJson().getString("name");
-        final Fruit entity = new Fruit(name);
-
-        return session.persist(entity)
-                        .onItem().produceUni(s -> session.flush())
-                        .onItem().apply(ignore -> httpResponse(rc, 201, entity));
+    public Uni<Fruit> create(@Body Fruit fruit, HttpServerResponse response) {
+        if (fruit == null || fruit.getId() != null) {
+            return Uni.createFrom().failure(new IllegalArgumentException("Fruit id invalidly set on request."));
+        }
+        return session.persist(fruit)
+                .onItem().transformToUni(session -> session.flush())
+                .onItem().transform(ignore -> {
+                    response.setStatusCode(201);
+                    return fruit;
+                });
     }
 
     @Route(methods = PUT, path = "/:id")
-    public Uni<HttpServerResponse> update(RoutingContext rc) {
-        final Integer id = Integer.valueOf(rc.request().getParam("id"));
-        final String name = rc.getBodyAsJson().getString("name");
-
-        return session.find(Fruit.class, id)
-                // if entity exists
-                .onItem().ifNotNull()
-                        .produceUni(entity -> {
-                                // Update the entity
-                                entity.setName(name);
-                                return session.flush()
-                                        .onItem().apply(ignore -> httpResponse(rc, 200, entity));
-                        })
-                // if entity doesn't exist
-                .onItem().ifNull()
-                        // Not found response code
-                        .continueWith(httpResponse(rc, 404));
+    public Uni<Fruit> update(@Body Fruit fruit, @Param String id) {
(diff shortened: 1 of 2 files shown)


mkouba:
CC @DavideD

cescoffier:
I just merged https://github.com/quarkusio/quarkus/pull/10871.

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
```diff
@@ -1,123 +1,112 @@
...
-        } catch (JsonProcessingException e) {
-            throw new IllegalArgumentException(e);
-        }
-    }
-
     @Route(methods = GET, path = "/")
-    public Multi<Fruit> getAll(RoutingContext rc) throws Exception {
-        return session.createNamedQuery(Fruit.FIND_ALL, Fruit.class).getResults();
+    public Multi<Fruit> getAll() throws Exception {
+        return ReactiveRoutes.asJsonArray(session.createNamedQuery(Fruit.FIND_ALL, Fruit.class).getResults());
```
DavideD, line 40: Is `ReactiveRoutes.asJsonArray(` required now?

mkouba: I think that it was required even before but the test does not validate the proper syntax of json array...

cescoffier: Yes, it required (For now) to creates an asynchronous json array. Each item is written in the response when ready (and sent to the client). 

We are improving this by inferring this from the `produces` attribute. But it's WIP...

DavideD: It surprised me because when I write
```
    @Route(methods = GET, path = "/")
    public Uni<List<Fruit>> getAll(RoutingContext rc) throws Exception {
        return session.createNamedQuery( Fruit.FIND_ALL, Fruit.class).getResultList();
    }
```

It works as expected. I'm wondering if it wouldn't look better to return a `Uni<List>`?

DavideD:
By the way, does the quickstart work when you go to http://localhost:8080/index.html?
I removed this dependency last minute thinking that it was unnecessary
```
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-resteasy-jsonb</artifactId>
        </dependency>
```
I forgot to test the web page and it doens't seem to work without it. Any idea why?

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
Sanne: yes it looks better, but the quickstart is supposed to teach to people "the right way".

@cescoffier are you optimistic about your WIP to be doable? If so, maybe it's ok for the example to already show the simpler syntax?  We're not working to win a benchmark war _yet_ ;)

DavideD: The right way is the one that works ;)

gsmet:
It would be weird that you would need any RESTEasy dependency when going full reactive routes. But I'll let Martin answer.

CI is failing with:
```
2020-07-21T13:57:20.3143319Z 2020-07-21 13:57:20,311 ERROR [org.acm.hib.rea.FruitsRoutes] (vert.x-eventloop-thread-2) Failed to handle request: java.lang.NullPointerException
2020-07-21T13:57:20.3144029Z 	at org.acme.hibernate.reactive.FruitsRoutes_RouteHandler_update_2420491368c662f8824cb079244c86a88df9203b.invoke(FruitsRoutes_RouteHandler_update_2420491368c662f8824cb079244c86a88df9203b.zig:104)
2020-07-21T13:57:20.3144426Z 	at io.quarkus.vertx.web.runtime.RouteHandler.handle(RouteHandler.java:53)
2020-07-21T13:57:20.3144773Z 	at io.quarkus.vertx.web.runtime.RouteHandler.handle(RouteHandler.java:17)
2020-07-21T13:57:20.3145132Z 	at io.vertx.ext.web.impl.RouteState.handleContext(RouteState.java:1034)
2020-07-21T13:57:20.3145481Z 	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:95)
2020-07-21T13:57:20.3145837Z 	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
2020-07-21T13:57:20.3146183Z 	at io.vertx.ext.web.handler.impl.BodyHandlerImpl.handle(BodyHandlerImpl.java:95)
2020-07-21T13:57:20.3146525Z 	at io.vertx.ext.web.handler.impl.BodyHandlerImpl.handle(BodyHandlerImpl.java:42)
2020-07-21T13:57:20.3146913Z 	at io.quarkus.vertx.http.runtime.VertxHttpRecorder$16.handle(VertxHttpRecorder.java:1053)
2020-07-21T13:57:20.3147279Z 	at io.quarkus.vertx.http.runtime.VertxHttpRecorder$16.handle(VertxHttpRecorder.java:1026)
2020-07-21T13:57:20.3147626Z 	at io.vertx.ext.web.impl.RouteState.handleContext(RouteState.java:1034)
2020-07-21T13:57:20.3148033Z 	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:131)
2020-07-21T13:57:20.3148369Z 	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
2020-07-21T13:57:20.3148716Z 	at io.vertx.ext.web.handler.impl.BodyHandlerImpl$BHandler.doEnd(BodyHandlerImpl.java:296)
2020-07-21T13:57:20.3149076Z 	at io.vertx.ext.web.handler.impl.BodyHandlerImpl$BHandler.end(BodyHandlerImpl.java:276)
2020-07-21T13:57:20.3149426Z 	at io.vertx.ext.web.handler.impl.BodyHandlerImpl.lambda$handle$0(BodyHandlerImpl.java:87)
2020-07-21T13:57:20.3149776Z 	at io.vertx.core.http.impl.HttpServerRequestImpl.onEnd(HttpServerRequestImpl.java:521)
2020-07-21T13:57:20.3150166Z 	at io.vertx.core.http.impl.HttpServerRequestImpl.lambda$pendingQueue$1(HttpServerRequestImpl.java:11
```
so there's something wrong going on.

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
cescoffier: Multi<X> and Uni<List<X>> don't have the same semantic. Multi<X> writes the X into the response when they are received, avoiding the items to be stored in memory. Uni<List<X>>> builds the list and write it in one "batch". We recommend List<X> is it fits in memory and if the items of the list can be retrieved in one _batch_. In which case are we with Hibernate Reactive?

cescoffier:
By removing the json dependency it removes the reasteasy extension and the index page is not served anymore.

gsmet:
You need RESTEasy to server static resources even when using Reactive routes?

cescoffier:
I believe so, maybe not. I can't remember if we moved to the Vert.x Static file handler or if it's still the Resteasy feature that is used.

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
DavideD: Hibernate Reactive at the moment only works with `Uni<List>` and the Multi returned by `query.getResults()` is just the Uni converted to a Multi.

That said, this quickstart is a simple case that returns a bunch of fruits, picking a Multi or a Uni in the API is more a question of what features we want to show.

The reason I used `Multi<Fruit>` is because it looks better and that was my goal for the quickstart.
But if it needs extra handling, I would return a Uni and add a comment to explain why it makes sense in this case to use it (instead of a Multi) and maybe create another example that shows when it makes sense to use a Multi.

mkouba:
@gsmet That's very likely a failure related to https://github.com/quarkusio/quarkus/pull/10871. It should be fixed now.

@DavideD @gsmet @cescoffier For resteasy standalone we register a wrapped `io.vertx.ext.web.handler.StaticHandler` here: https://github.com/quarkusio/quarkus/blob/master/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/ResteasyStandaloneRecorder.java#L134-L151  (by the way the original vertx issue was fixed in 3.8.4 so it should not be necessary to wrap it anymore). We should probably move the static handler logic to `vertx-http`?

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
mkouba: `Uni<List<Fruit>>` works fine and makes probably more sense for this use case (unless there's a plan to support `Multi` "properly").

@cescoffier @Sanne Should we change the return type to `Uni<List<Fruit>>`? And if so what kind of comment should we add? _"Hibernate Reactive does not support this yet"_ does not sound good ;-).

cescoffier: From a reactive standpoint, returning a Multi is better.

cescoffier:
@mkouba Yes, we should move this to vertx-http. @stuartwdouglas What's your opinion on this?

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
cescoffier: As I said, there are also good arguments to return a `Uni<List>` so I don't mind.

DavideD:
Out of curiosity, what's the difference between `vertx-web` and `vertx-http`?

mkouba:
> Out of curiosity, what's the difference between vertx-web and vertx-http?

`vertx-web` = reactive routes, user-facing API. `vertx-http` - core components around the web server.

DavideD:
I see, thanks

stuartwdouglas:
Sounds reasonable.

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
Sanne: thanks @cescoffier , let's use what looks best then.

> Should we change the return type to `Uni<List<Fruit>>`? And if so what kind of comment should we add? _"Hibernate Reactive does not support this yet"_ does not sound good ;-).

I don't feel a comment is necessary. What is triggered when using this API is a single SQL statement which expects a reasonably sized result set. For example, here we'll load all of 4 fruits from the DB.. there's no need to use a stream for this, in fact it would be less efficient as it's more chatty to the DB, while the whole resultset is likely to fit in a single network packet.

For streaming results, Hibernate offers different APIs which people are encouraged to use when the expected size of the results is significant. (while getResultsList() is intentionally returning a `Uni<List<R>>` ).

Sanne: P.S. I see that the alternative API which I referred to - one which produces a "real Multi" - isn't around yet, but I expect we'll be able to make one based on the scrollable API of ORM - wich is backed by server side cursors.

Sanne: @gavinking ^

mkouba: > let's use what looks best then.

`Uni<List<Fruit>>` then? ;-)

DavideD: > "Hibernate Reactive does not support this yet" does not sound good 

I don't think this is a Hibernate Reactive issue, it's a general issue when designing the API. Not saying that we need to add a comment but something similar to this would tell the user the differences:
```
In this case it makes sense to return a Uni<List<Fruit>> because we return a reasonable amount of results. Consider return a Multi<Fruit> for streaming results (link to example).
```

DavideD: For everybody interested, this is the issue explaining why Hibernate Reactive hasn't a proper Multi return method: https://github.com/hibernate/hibernate-reactive/issues/153

Having some feedback is always nice.

mkouba:
The CI failure is probably unrelated:
```
2020-07-24T11:15:15.8660015Z [ERROR]   AggregatorTest.test ? Timeout test() timed out after 10 seconds
...
2020-07-24T11:15:15.8788004Z [INFO] kafka-streams-quickstart-aggregator ................ FAILURE [ 44.550 s]
```

mkouba:
@DavideD The CI is green now! Feel free to merge this PR ;-).

gsmet:
I don't think he can :).

Let me know if it's ready to go in, I will merge it.

mkouba:
> I don't think he can :).

Ah, you're right. So I think it's ready but maybe one more +1 would be good ;-).

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
```diff
@@ -1,123 +1,113 @@
...
-                                return session.flush()
-                                        .onItem().apply(ignore -> httpResponse(rc, 200, entity));
-                        })
-                // if entity doesn't exist
-                .onItem().ifNull()
-                        // Not found response code
-                        .continueWith(httpResponse(rc, 404));
+    public Uni<Fruit> update(@Body Fruit fruit, @Param String id) {
+        if (fruit == null || fruit.getName() == null) {
+            return Uni.createFrom().failure(new IllegalArgumentException("Fruit name was not set on request."));
```
DavideD, line 65: A curiosity, does it have to be a Uni in this case? Can't we just throw the exception?

DavideD:
It's OK for me to merge it (somebody else needs to do ti though) but without the additional dependency going to http://localhost:8080/index.html is not going to work (unless something has changed already). We should also add a test that checks that the page is up and running so that we won't forget about it in the future.

mkouba:
> but without the additional dependency going to http://localhost:8080/index.html is not going to work (unless something has changed already).

This is already fixed in master...

hibernate-reactive-routes-quickstart/src/main/java/org/acme/hibernate/reactive/FruitsRoutes.java
mkouba: We can. But I think that it's better to avoid throwing an exception unless we have to...

gsmet:
Merging then, thanks everyone!

PR #634, opened 2020-08-03
Replace kscript with jbang for cron job

geoand:
This also needs to be cherry-picked into `master` in order for it to have any effect


.github/NativeBuildReport.java
```diff
@@ -0,0 +1,95 @@
+//usr/bin/env jbang "$0" "$@" ; exit $?
```
geoand, line 1: This was copied from the script we have in Quarkus proper

gastaldi: We should move this class to https://github.com/quarkusio/jbang-catalog at some point

.github/workflows/native-build-development.yml
```diff
@@ -86,13 +86,12 @@ jobs:
...
             -o Dir::Etc::sourceparts="-" -o APT::Get::List-Cleanup="0"
           sudo apt-get install -y gnupg2 gnupg-agent
           echo Installing SDKMAN
           curl -s "https://get.sdkman.io" | bash
           source ~/.sdkman/bin/sdkman-init.sh && \
-          sdk install kotlin 1.3.61 && \
-          sdk install kscript 2.9.0
-          echo "The report step got status: ${STATUS}"
-          kscript .github/NativeBuildReport.kts ${GITHUB_TOKEN} ${STATUS}
+          sdk install jbang 0.36.1
```
gastaldi, line 96: There is a `jbangdev/jbang-action@v0.36.1` Github action which could be used here instead
eg.
```yaml
      - name: Index Catalog
        uses: jbangdev/jbang-action@v0.36.1
        with:
          script: indexcatalog@quarkusio
          args: . ./app/registry.json
          trust: https://github.com/quarkusio/jbang-catalog
```

geoand:
@gastaldi great suggestions!

How about we (as in someone that is not me since I'll be off for a couple of weeks  ☺️) do them as a follow-up?

.github/NativeBuildReport.java
maxandersen: A central shortcut catalog is not suitable for hosting scripts to do releases. You want the release scripts tag/versioned together with your artifacts.

You can add a catalog file to this Repo if you want to simplify its usage but I would highly recommend to not move these scripts to central catalog.

PR #635, opened 2020-08-03
Avoid building GCF native image

loicmathieu:
**Check list**:

Your pull request:

- [x] targets the `development` branch
- [x] uses the `999-SNAPSHOT` version of Quarkus
- [ ] has tests (`mvn clean test`)
- [ ] works in native (`mvn clean package -Pnative`)
- [ ] has native tests (`mvn clean verify -Pnative`)
- [x] makes sure the documentation must not be updated
- [ ] links the documentation update pull request (if needed)
- [ ] updates or creates the `README.md` file (with build and run instructions)
- [ ] For new quickstart, is located in the directory _component-quickstart_
- [ ] For new quickstart, is added to the root `pom.xml` and `README.md`

PR #639, opened 2020-08-10
Funqy Google Cloud Functions quickstart

loicmathieu:
**Check list**:

Your pull request:

- [x] links the documentation update pull request (if needed)
- [x] For new quickstart, is located in the directory _component-quickstart_
- [x] For new quickstart, is added to the root `pom.xml` and `README.md`


loicmathieu:
The related document PR is here: https://github.com/quarkusio/quarkus/pull/11298

loicmathieu:
No tests and no native support for Google Cloud Functions.
If the PR #635 si merged first this one will need to be updated to avoid building it in native.
If not, I will update #635 to also avoid building the new quickstart.

gsmet:
I disabled the native build. That being said, I wonder if we should instead remove the profile from the pom?

loicmathieu:
@gsmet removing the profile will make the integration test builds again in native but without building the native image and executing the test twice right ?
If yes, this is way better as it allow to control from the extension the exclusion from native image compilation instead of maintaining a list of excluded projects ...

gsmet:
No, if you remove the profile altogether, it won't run anything native and just run the tests in JVM mode (as long as there are no native tests, but I suppose we removed them, right?).

loicmathieu:
OK, I'll give it a try, it should be better for both quickstarts and extensions build.

PR #649, opened 2020-09-02
Make the JMS quickstart test more robust

cescoffier:
@geoand That should fix the issue you reported this morning.

In addition, using a Thread.sleep is everything but reliable...


geoand (approved):
Makes perfect sense :)

PR #654, opened 2020-09-09
Fix the Hibernate Reactive pom files

gsmet:
These things might have been necessary at some point if the bom was incorrect but they shouldn't be.

/cc @DavideD @cescoffier


cescoffier (approved):
LGTM! Let's wait for CI.

DavideD:
Nice!

PR #655, opened 2020-09-09
Remove tests for the exception messages in hibernate-reactive-* quick…

DavideD:
@gsmet I've removed the tests causing the exceptions on the console.

Surprisingly, I don't see strange logs anymore. I don't know what happened but don't think it's related


gsmet:
Perfect, thanks!

PR #673, opened 2020-09-30
Skip tests for Amazon KMS

gsmet:
@marcinczeczko I don't know what's happening but the Amazon KMS quickstart started to fail with a timeout.

Any chance you could have a look? It might be related to a dependency upgrade.


marcinczeczko:
@gsmet - no idea why it stopped working :P however, I upgraded couple bits and pieces and the KMS port in #674  and apparently it works now - let see in CI how it performs.

PR #683, opened 2020-10-05
Another attempt to try to understand what's going on with the JMS quickstart in native

cescoffier:
The native build gets 204 response instead of 200 response, showing that the message is not received from JMS.
A message is sent every 5 seconds, but do not seem to be received. Trying to understand why...

Of course, can't reproduce locally.

### DIFF
--- a/jms-quickstart/src/main/java/org/acme/jms/PriceConsumer.java
+++ b/jms-quickstart/src/main/java/org/acme/jms/PriceConsumer.java
@@ -52,6 +52,7 @@ public class PriceConsumer implements Runnable {
                     return;
                 }
                 lastPrice = message.getBody(String.class);
+                System.out.println("Got last price: " + lastPrice);
             }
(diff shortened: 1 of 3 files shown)


gastaldi (approved):

jms-quickstart/src/main/java/org/acme/jms/PriceConsumer.java
```diff
@@ -52,6 +52,7 @@ public void run() {
                     return;
                 }
                 lastPrice = message.getBody(String.class);
+                System.out.println("Got last price: " + lastPrice);
```
gastaldi, line 55: Use log.info?

cescoffier: There are no logger. We already have System.out in the test resource, so discarded the use of it.
Hopefully, once I understand the issue, this will be reverted.

PR #687, opened 2020-10-05
Migrate Kafka Streams quickstart code to Mutiny

alexey-anufriev:
Related issue #12520

### DIFF
--- a/kafka-streams-quickstart/producer/src/main/java/org/acme/kafka/streams/producer/generator/ValuesGenerator.java
+++ b/kafka-streams-quickstart/producer/src/main/java/org/acme/kafka/streams/producer/generator/ValuesGenerator.java
@@ -2,21 +2,19 @@ package org.acme.kafka.streams.producer.generator;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Random;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
 
+import io.smallrye.mutiny.Multi;
+import io.smallrye.reactive.messaging.kafka.Record;
 import org.eclipse.microprofile.reactive.messaging.Outgoing;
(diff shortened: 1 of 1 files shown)


kafka-streams-quickstart/producer/src/main/java/org/acme/kafka/streams/producer/generator/ValuesGenerator.java
```diff
@@ -60,13 +60,13 @@
     }
 
     @Outgoing("weather-stations")
-    public Flowable<KafkaRecord<Integer, String>> weatherStations() {
+    public Multi<KafkaRecord<Integer, String>> weatherStations() {
         List<KafkaRecord<Integer, String>> stationsAsJson = stations.stream()
```
cescoffier, line 64: Minor - 2 things:

* we can create a Multi directly from a Stream (`Multi.createFrom().items(stream)`), it will iterate according to demands, which is often better
* I don't believe we need to use `KafkaRecord`, we can use `Record` directly (depends who consume the stream).

Can you check with:

```java
public Multi<Record<Integer, String>> weatherStations() {
   return Multi.createFrom().items(stations.stream()
      .map(s -> Record.of(s.id, "{ \"id\" : " + s.id + ", \"name\" : \"" + s.name + "\" }"))
   );
}
```

cescoffier (changes requested):

kafka-streams-quickstart/producer/src/main/java/org/acme/kafka/streams/producer/generator/ValuesGenerator.java
alexey-anufriev: @cescoffier, could you please let me know which `Record` you mean in this case? Did not find this class right away.

cescoffier: This one: `io.smallrye.reactive.messaging.kafka.Record`. It's quite new, so be sure you have the latest Quarkus _master_.

alexey-anufriev:
Updated implementation according to the suggestion.

cescoffier (approved):

cescoffier:
Thanks!

PR #696, opened 2020-10-19
Adding step on the `getting-started` project to run quarkus on non-linux systems natively

rafadelnero:
For developers who are starting using `Quarkus`, it's very important for them to have the example working. 

The README file from the `getting-started` project only covers the command to run `Quarkus` natively on Linux but the command to run natively on non-Linux systems is not there.

To make the life of developers who are getting started with `Quarkus` I added the extra command to make it work. 

**Check list**:

Your pull request:

- [X ] targets the `development` branch
- [X ] has tests (`mvn clean test`)
- [ X] works in native (`mvn clean package -Pnative`)
- [ X] has native tests (`mvn clean verify -Pnative`)
- [ X] updates or creates the `README.md` file (with build and run instructions)

### DIFF
--- a/getting-started/README.md
+++ b/getting-started/README.md
@@ -72,6 +72,10 @@ Compiling a native executable takes a bit longer, as GraalVM performs additional
 steps to remove unnecessary codepaths. Use the  `native` profile to compile a
 native executable:
 
+To run on a non-Linux system (GraalVM and Docker required):
+> ./mvnw clean install -Pnative -Dnative-image.docker-build=true
+
+To run on Linux:
 > ./mvnw install -Dnative
 
 After getting a cup of coffee, you'll be able to run this executable directly:


getting-started/README.md
```diff
@@ -72,6 +72,10 @@ Compiling a native executable takes a bit longer, as GraalVM performs additional
 steps to remove unnecessary codepaths. Use the  `native` profile to compile a
 native executable:
 
+To run on a non-Linux system (GraalVM and Docker required):
+> ./mvnw clean install -Pnative -Dnative-image.docker-build=true
```
cescoffier, line 76: It's not totally true.

```
./mvnw install -Dnative
```

works on all systems, but the resulting executable is specific to the host. So except on Linux, you cannot copy that file into a container. 

I would extend the documentation and add a container section.

cescoffier (changes requested):

getting-started/README.md
gsmet: /me still thinking having these READMEs there is a bad idea and they should simply point to the guide where we would centralize everything.

cescoffier: I agree! I can't remember why we added them at the beginning (maybe it was before we had any doc... good old times :-D )

rafadelnero: Thanks for answering @cescoffier and @gsmet! Apologies for the delay in my answer.  

I also appreciate the clarification regarding the following command:
`./mvnw install -Dnative`

Extending to the documentation to add a container section seems the best option. I can take a further look at the Quarks documentation.

bartwinski: Regarding building with ./mvnw install -Dnative, it is not compiling (see my post https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Native.20build.20via.20Docker.20fails/near/219348056). I got Ubuntu 20.04 and jdk11. The same compilation problems occurs with only ./mvnw install -Dnative.

rafadelnero: Interesting, thanks for the comment @bartwinski! Should new Quarkus users need to go deep into those matters or should Quarkus have a command that works right away? Maybe adding more detailed documentation in the README file would be useful. 

Alternatively, adding a command that just works would make other developer's life easier.

cescoffier:
@gsmet what's your opinion on this one?

PR #711, opened 2020-11-13
Use Jackson in the quickstarts

gsmet:
Let's see what CI has to say about this change.

I'm preparing the doc counterparts.

### DIFF
--- a/mongodb-panache-quickstart/src/main/java/org/acme/mongodb/panache/Person.java
+++ b/mongodb-panache-quickstart/src/main/java/org/acme/mongodb/panache/Person.java
@@ -5,11 +5,14 @@ import java.util.List;
 
 import org.bson.codecs.pojo.annotations.BsonProperty;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import io.quarkus.mongodb.panache.MongoEntity;
 import io.quarkus.mongodb.panache.PanacheMongoEntity;
 
 @MongoEntity(collection = "ThePerson")
 public class Person extends PanacheMongoEntity {
+    @JsonProperty
     public String name;
     // will be persisted as a 'birth' field in MongoDB
     @BsonProperty("birth")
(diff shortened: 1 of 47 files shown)


mongodb-panache-quickstart/src/main/java/org/acme/mongodb/panache/Person.java
```diff
@@ -5,11 +5,14 @@
...
 import org.bson.codecs.pojo.annotations.BsonProperty;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import io.quarkus.mongodb.panache.MongoEntity;
 import io.quarkus.mongodb.panache.PanacheMongoEntity;
 
 @MongoEntity(collection = "ThePerson")
 public class Person extends PanacheMongoEntity {
+    @JsonProperty
```
gsmet, line 15: @loicmathieu I'm not really happy about this change but, when switching to Jackson, I end up with `name` being null in the deserialized `Person`.

loicmathieu: It's strange.
I remember noticing that Jackson and JSON-B are not configured the same way. And Restassured used a different ObjectMapper configuration than the application.

Anyway, your change is strange, it should not have any impact and `name` should not be null.
If you want to merge your PR with this change, create a followup issue so someone (me or @geoand, the master of serialization ;) ) can have a look.

geoand: Yeah, I had a PR to make RestAssured use the applications's `ObjectMapper`, but I never merged it because it does not work in native tests and would therefore create far too much confusion.

gsmet: so `name` is serialized OK by Quarkus but somehow the ObjectMapper of REST Assured doesn't deserialize it properly. What is extremely weird is that REST Assured was already using Jackson.

PR #726, opened 2020-12-11
Fix mongodb test

loicmathieu:
**Check list**:

Your pull request:

### DIFF
--- a/mongodb-panache-quickstart/src/test/java/org/acme/mongodb/panache/PersonResourceTest.java
+++ b/mongodb-panache-quickstart/src/test/java/org/acme/mongodb/panache/PersonResourceTest.java
@@ -97,7 +97,7 @@ class PersonResourceTest {
                 .body().as(Person.class);
 
         Assertions.assertThat(person.id).isEqualTo(new ObjectId("5889273c093d1c3e614bf2fa"));
-        Assertions.assertThat(person.name).isEqualTo("LOÏC");
+        Assertions.assertThat(person.name).isEqualTo("loïc");
         Assertions.assertThat(person.birthDate).isEqualTo(LocalDate.of(1988, 6, 19));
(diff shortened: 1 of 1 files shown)


loicmathieu:
/cc @geoand 

Fixes the current CI issue for MongoDB with Panache

geoand (approved):

geoand:
Thanks!

rsvoboda:
@loicmathieu what change in Quarkus master caused this change? I was just looking into mongodb-panache-quickstart fail.

loicmathieu:
@rsvoboda accessors rewritting as exist for Hibernate with Panache has been implemented, so now person.name was rewritten to person.getName() inside the Panache processor and the test fail.

rsvoboda:
ok, so https://github.com/quarkusio/quarkus-quickstarts/blob/master/mongodb-panache-quickstart/src/main/java/org/acme/mongodb/panache/Person.java#L23-L25 is called now, instead of getting the direct property value.

Should be a note about this change of behavior added into https://github.com/quarkusio/quarkus/wiki/Migration-Guide-1.11?

loicmathieu:
@evanchooly the just merged field accessor rewritting you implemented for MongoDB with Panache can cause some bugs on existing code where getters/setters exists but developer still use the field directly. This was the case on the quickstart.

As @rsvoboda suggested, maybe it'll be a good idea to add a note about it inside the migration guide for 1.11

evanchooly:
wouldn't a simple rebuild fix this?  I can certainly add a note to that effect but I'm a little surprised it broke a test since rebuilding them should have updated the bytecode accordingly.  What did the breakage look like?

loicmathieu:
There was a lowercase on the getter, previously it was not used, now it is.

evanchooly:
oh!  i see.  That entity had hand crafted a getter which was bypassed with field access.  Got it.  That was the missing piece for me.

PR #733, opened 2020-12-18
Revert "Use non application endpoints prefix introduced in #13601"

rsvoboda:
This reverts commit 9c802b2f0f985644dba23497886fb24ca9baab47.

Redirect of old URLs should work and it was fixed in Quarkus upstream.
Tracked in https://github.com/quarkusio/quarkus/issues/13901


gsmet (changes requested):
I think it's better to use the new URL so I don't think we should merge this.

rsvoboda:
@gsmet, ok, closing

PR #751, opened 2021-01-19
Fix MailerResource as MailTemplateInstance#send method returns Uni

rsvoboda:
Fix MailerResource as MailTemplateInstance#send method returns Uni

Change introduced in https://github.com/quarkusio/quarkus/pull/14344
https://quarkus.io/guides/mailer doesn't need any change, already shows  `.subscribeAsCompletionStage()`

### DIFF
--- a/mailer-quickstart/src/main/java/org/acme/getting/started/MailerResource.java
+++ b/mailer-quickstart/src/main/java/org/acme/getting/started/MailerResource.java
@@ -1,6 +1,5 @@
 package org.acme.getting.started;
 
-import javax.inject.Inject;
 import javax.validation.Valid;
 import javax.validation.constraints.Email;
 import javax.validation.constraints.NotBlank;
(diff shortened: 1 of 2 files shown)


rsvoboda:
Should be merged after 1.11.0.Final tag is done as the mailer change is for 1.12

mailer-quickstart/src/main/java/org/acme/getting/started/MailerResource.java
```diff
@@ -35,6 +35,7 @@
                 .to(email)
                 .subject("Ahoy " + name + "!")
                 .send()
+                .subscribeAsCompletionStage()
```
cescoffier, line 38: It would be better to change the signature and return a Uni. This is the model we recommend.

cescoffier (changes requested):

mailer-quickstart/src/main/java/org/acme/getting/started/MailerResource.java
rsvoboda: So the docs need update too then?
https://quarkus.io/guides/mailer - `public CompletionStage<Response>`

cescoffier: yes

rsvoboda:
@cescoffier PR updated

cescoffier (approved):
Thanks @rsvoboda

rsvoboda:
About https://quarkus.io/guides/mailer update, I don't really feel confident to do appropriate changes.
I expect few things about Uni vs. CS should be mentioned in the guide to explain why Uni is the way Quarkus focuses on.

cescoffier:
@rsvoboda Can you open an issue in Quarkus and assigned it to me. I can't promise I will provide a PR this week, but should not be too far in the future.

gsmet:
Can you revert please? We need to wait for 1.11 to be released as stated by @rsvoboda above.

cescoffier:
My bad!

PR #760, opened 2021-01-22
Use RESTEasy Reactive in getting-started-reactive-crud quickstart

geoand:


getting-started-reactive-crud/src/main/java/org/acme/reactive/crud/FruitResource.java
```diff
@@ -90,15 +77,15 @@ private void initdb() {
 
     @PUT
     @Path("{id}")
-    public Uni<Response> update(@PathParam Long id, Fruit fruit) {
+    public Uni<Response> update(Long id, Fruit fruit) {
```
cescoffier, line 80: oh oh! I like this feature! I was not aware of it :-D

cescoffier (approved):

getting-started-reactive-crud/src/main/java/org/acme/reactive/crud/FruitResource.java
geoand: It's pretty useful indeed :).

Basically when the names match, you don't need the annotation

PR #763, opened 2021-01-27
Don't specify the native image builder

geoand:
Without this, the cron job no longer works
and https://github.com/quarkusio/quarkus/issues/6588 is filled with errors


cescoffier (approved):

geoand:
It seems like needs to be cherry-picked to `master` in order for the cron job to take it into account

PR #767, opened 2021-01-29
Use 1.0.0-SNAPSHOT as version in all quickstarts

ia3andy:
(cherry picked from commit 1f00d63992d2a44121c81a6e3a40673384bb9e76)

cc @gsmet

### DIFF
--- a/google-cloud-functions-http-quickstart/src/main/resources/META-INF/resources/index.html
+++ b/google-cloud-functions-http-quickstart/src/main/resources/META-INF/resources/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <title>google-cloud-functions-http - 1.0-SNAPSHOT</title>
+    <title>google-cloud-functions-http - 1.0.0-SNAPSHOT</title>
     <style>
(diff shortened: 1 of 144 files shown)


google-cloud-functions-http-quickstart/src/main/resources/META-INF/resources/index.html
```diff
@@ -135,7 +135,7 @@ <h3>Application</h3>
             <ul>
                 <li>GroupId: org.acme</li>
                 <li>ArtifactId: google-cloud-functions-http</li>
-                <li>Version: 1.0-SNAPSHOT</li>
+                <li>Version: 1.0.0-SNAPSHOT</li>
                 <li>Quarkus Version: 1.6.1.Final</li>
```
ia3andy, line 139: @gsmet is that normal `1.6.1.Final`?

gsmet (approved):

google-cloud-functions-http-quickstart/src/main/resources/META-INF/resources/index.html
gsmet: I think they are updated when pushed to the master branch.

PR #768, opened 2021-01-29
Explicitly add the rest-client-jackson extension

gsmet:
People are trying this quickstart with RESTEasy Reactive and if they
switch to it, it's missing the RESTEasy classic Jackson support.


geoand (approved):
Makes perfect sense!

PR #777, opened 2021-02-10
Add kafka-panache-quickstart into root pom.xml

rsvoboda:
Add kafka-panache-quickstart into root pom.xml


rsvoboda:
@gsmet if you could merge this? And also add it into 1.11.3.Final once you work on the tag.

We would like to run this QS as part of RHBQ 1.11 testing.

gsmet (approved):

PR #781, opened 2021-02-16
Don't run CI in forks by default

famod:
IMO it is rather confusing to have a cron scheduled job running in a fork for a branch that you will most likely never push to directly.

This PR auto-skips this job in forks but at the same time adds this nice feature:

grafik

So you can (in your fork) run tests on your yet to be merged branch.

PS: "Report" step is never run in forks for obvious reasons.

### DIFF
--- a/.github/workflows/native-build-development.yml
+++ b/.github/workflows/native-build-development.yml
@@ -6,14 +6,16 @@ on:
   schedule:
     - cron: '0 2 * * *'
   repository_dispatch:
+  workflow_dispatch:
 
 jobs:
   build_with_native:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/native-build-development.yml
```diff
@@ -6,14 +6,16 @@ on:
...
 
 jobs:
   build_with_native:
     runs-on: ubuntu-latest
+    if: "github.repository == 'quarkusio/quarkus-quickstarts' || github.event_name == 'workflow_dispatch'"
     steps:
       - uses: actions/checkout@v2
         with:
-          ref: development
+          ref: ${{ github.event_name == 'workflow_dispatch' && github.event.ref || 'development' }}
```
famod, line 18: Note: This is a "fake ternary" as a workaround for missing native support by GH.

gsmet (approved):

PR #782, opened 2021-02-16
Align 'Build Quarkus master' to ci-actions.yml

famod:
See https://github.com/quarkusio/quarkus/blob/master/.github/workflows/ci-actions.yml#L94

Instead of _some_ `-DskipX` properties we can go the full monty with `-Dquickly-ci`.


famod:
Btw, we could save some time by skipping all `integration-tests` but that would require another profile.

gsmet (approved):

PR #783, opened 2021-02-17
Fix quarkus plugin and platform versions (missing dependency in maven).

ibalashov:
Fixing maven dependency. It contradicts the two checklist items below, not sure how relevant they are here.


gsmet (approved):
Thanks. The module was missing in the pom, that's why the version was not updated. It has been fixed in `development` so this is OK.

PR #784, opened 2021-02-18
Introduce config-yaml-quickstart

famod:
Should make issues like https://github.com/quarkusio/quarkus/issues/15173 less likely.
It's basically a copy of `config-quickstart`, just with YAML.

**Check list**:

Your pull request:

- [x] has native tests (`mvn clean verify -Pnative`)
- [x] makes sure the associated guide must not be updated
- [x] for new quickstart, is located in the directory _component-quickstart_
- [x] for new quickstart, is added to the root `pom.xml` and `README.md`


config-yaml-quickstart/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy</artifactId>
```
geoand, line 34: I propose the use of `quarkus-resteasy-reactive` :)

famod: What a surprise! 😉 

But isn't this a general topic for almost all _quickstarts_ (separate PR)?
If we do this here the other config quickstart should be adjusted as well, no?

geoand: I'd say that let's do it for new ones. At some point we will certainly change the rest as well

famod: I think it's important to have a comparability with the other config quickstart, so I will change both.

famod: Done. I don't think those simple methods need `@Blocking` (reactive newbie here).

famod: #786 tracks the general move for the rest.

gsmet:
Frankly, I'm not sure I see the value of this. And I also don't see how it will prevent people from using 4 spaces instead of 2.

config-yaml-quickstart/pom.xml
geoand: > Done. I don't think those simple methods need `@Blocking` (reactive newbie here).

Yup, needed `@Blocking` for this case

famod:
It cannot prevent anything, but since (new) users often seem to have a look at the quickstarts first and YAML is rather popular, I thought it's a good idea to have one for YAML config.
No other quickstart is using YAML config.

gsmet:
Yeah. But it's one more thing to build and we start to be tight on the native-image CI front for the quickstarts.

Maybe if we drop the native part from it. And if we reference it properly in the doc.

famod:
> we start to be tight on the native-image CI front for the quickstarts

Ah, ok - didn't know that!

> Maybe if we drop the native part from it

Sure, I can do that. I was initially thinking about just a YAML Maven profile in `config-quickstart` but didn't do that in the end because it might be confusing...?

> And if we reference it properly in the doc.

What is "it" exactly? Sorry, long day. 😉

gsmet:
Should have been more clear :). We need a pointer in the doc pointing to the quickstart. Otherwise it won't get noticed.

Also, I think it would be worth adding a warning in the doc about the 2 spaces constraint.

famod:
Well, what can I say...this whole spaces problem was a phantom: https://github.com/quarkusio/quarkus/issues/15173#issuecomment-782341212

I suppose this makes this PR even less viable.

PS: On the plus side, this PR led to #785 and #786. :wink:

PR #800, opened 2021-03-03
Move quickstarts to use DevServices

stuartwdouglas:
Some more complex ones still use the docker
plugin, however this moves the majority to
use DevServices.

**Check list**:

Your pull request:


maxandersen:
i like the idea and maybe im blind but how does a user get going that does not  have docker running locally ?

i guess we can say it is the same as before, but it made me realize we aren't remembering all the users on company locked down machines without access to docker.

stuartwdouglas:
We could just add comment to remove the %prod. if you want to use a local database. If you are using a local DB you would need to configure it correctly anyway, so its not really that big a change.

gsmet (approved):

PR #804, opened 2021-03-03
Fix mongo quickstart

edtbl76:
**Check list**:

Your pull request:

- [X] targets the `development` branch
- [X] uses the `999-SNAPSHOT` version of Quarkus
- [X] has tests (`mvn clean test`)
- [X] works in native (`mvn clean package -Pnative`)
- [X] has native tests (`mvn clean verify -Pnative`)
- [X] makes sure the associated guide must not be updated
- [X] for new quickstart, is located in the directory _component-quickstart_
- [X] for new quickstart, is added to the root `pom.xml` and `README.md`

I didn't update or create a README.md for this. 

Let me know if you want to add one. I can copy/paste one from one of the other modules. 
I fixed the module and added tests, so now it works.

### DIFF
--- a/amazon-dynamodb-quickstart/README.md
+++ b/amazon-dynamodb-quickstart/README.md
@@ -5,7 +5,11 @@ This example showcases how to use the AWS DynamoDB client with Quarkus. As a pre
 # DynamoDB local instance
 
 Just run it as follows:
-`docker run --rm --name local-dynamo -p 8000:4569 -e SERVICES=dynamodb -e START_WEB=0 -d localstack/localstack`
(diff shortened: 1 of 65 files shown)


edtbl76:
I just saw that you are solving this a different way using

edtbl76:
I noticed you are solving this a different way in the mongo-panache-quickstart. 

I'm assuming it make sense to have them both use the same mechanism? 

I was following the convention used in the AWS tests for the mongo-quickstart, I can modify panache to match, or modify mongo-quickstart to match panache. 

This would allow the mongo-panache module to use the actual mongo container. (Although your wrapper is a bit more flexible, because we can access the client in the test class in order to perform per-test setup/teardown)

PR #809, opened 2021-03-09
Hibernate reactive panache

DavideD:
A quickstart using Panache with Hibernate Reactive.
Note that I'm flushing the CRUD operations manually but I'm wondering if there is a way to support the `@Transactional` annotation.

Anyway, it seems to work and it should help people wanting to play with the experimental extension.

By the way, all the quickstarts are using Postgres 10.5, shouldn't we upgrade to 13?

Cheers


hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/Fruit.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.persistence.Entity;
+
+import io.quarkus.hibernate.reactive.panache.PanacheEntity;
+
+@Entity
+@Cacheable
+public class Fruit extends PanacheEntity {
+
+    @Column(length = 40, unique = true)
+    private String name;
```
FroMage, line 14: This should be a public field with no accessors: they're generated.

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
```diff
@@ -0,0 +1,136 @@
...
+                    } )
+                // else
+                .onItem().ifNull()
+                    .continueWith( Response.ok().status( 404 ).build() );
+    }
+
+    @DELETE
+    @Path("{id}")
+    public Uni<Response> delete(@RestPath Long id) {
+        return Fruit.findById(id)
```
FroMage, line 78: We have `deleteById`, would that be better?

README.md
```diff
@@ -44,6 +44,7 @@ See [CONTRIBUTING](CONTRIBUTING.md) for how to build these examples.
 * [Amazon S3](./amazon-s3-quickstart): How to connect to an Amazon S3 cloud storage
 * [Amazon SNS](./amazon-sns-quickstart): How to connect to an Amazon SNS (Simple Messaging Service)
 * [Amazon SQS](./amazon-sqs-quickstart): How to connect to an Amazon SQS (Simple Queue Service)
+* [Hibernate Reactive Panache and RESTEasy](./hibernate-reactive-quickstart): Exposing a CRUD service over REST using Hibernate Reactive and Panache to connect to a PostgreSQL database
```
FroMage, line 47:
```suggestion
* [Hibernate Reactive Panache and RESTEasy Reactive](./hibernate-reactive-quickstart): Exposing a CRUD service over REST using Hibernate Reactive and Panache to connect to a PostgreSQL database
```

hibernate-reactive-panache-quickstart/README.md
```diff
@@ -0,0 +1,118 @@
+# Quarkus demo: Hibernate Reactive with Panache and RESTEasy
```
FroMage, line 1:
```suggestion
# Quarkus demo: Hibernate Reactive with Panache and RESTEasy Reactive
```

hibernate-reactive-panache-quickstart/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+This is a minimal CRUD service exposing a couple of endpoints over REST,
+with a front-end based on Angular so you can play with it from your browser.
+
+While the code is surprisingly simple, under the hood this is using:
+ - RESTEasy to expose the REST endpoints
```
FroMage, line 7:
```suggestion
 - RESTEasy Reactive to expose the REST endpoints
```

FroMage (changes requested):
LGTM except minor comments

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
DavideD: Is there a way to flush without having the entity?

FroMage: https://github.com/quarkusio/quarkus/issues/15595

DavideD: What I need here is something like `persistAndFlush`. It seems there is no  equivalent for the delete.
I could change the code to something like this:
```
    @DELETE
    @Path("{id}")
    public Uni<Response> delete(@RestPath Long id) {
        return Fruit.deleteAndFlushById(id)
                .onItem().transform( deleted -> deleted
                        ? Response.ok().status(204).build()
                        : Response.ok().status(404).build()
        );
    }
```

FroMage: Good point. I'll add it to the issue.

FroMage (approved):
Great, thanks!

FroMage:
> What I'm wondering now if it would be better to wrap everything in a transaction using Panache.withTransaction instead of using flush. What do you think?

Hey, you're the DB expert, I defer to you ;)

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
DavideD: > Hey, you're the DB expert, I defer to you ;)

:) Yeah, I think we need to show how to create transactions

DavideD:
OK, I've updated the example and now it uses transactions.
I think it looks pretty good now.

Thanks

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @POST
+    public Uni<Response> create(Fruit fruit) {
+        if (fruit == null || fruit.id != null) {
+            throw new WebApplicationException("Id was invalidly set on request.", 422);
+        }
+
+        return Panache.withTransaction(fruit::persist)
+                .onItem().transform(ignore -> Response.ok(fruit).status(CREATED).build());
+    }
```
gavinking, line 61: @DavideD `onItem().transform()` is a rather verbose way to do this. I think it's better to use the new methods we added to Mutiny.

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
```diff
@@ -0,0 +1,143 @@
...
+        if (fruit == null || fruit.name == null) {
+            throw new WebApplicationException("Fruit name was not set on request.", 422);
+        }
+
+        return Panache
+                .withTransaction(() -> Fruit.<Fruit> findById(id)
+                        .onItem().ifNotNull().transform(entity -> {
+                            entity.name = fruit.name;
+                            return entity;
+                        }))
```
gavinking, line 75: Should use `invoke()` here.

gavinking (changes requested):
Looks to me like this code example is using all the old Mutiny APIs, not the new stuff we added to make this sort of code look prettier.

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
DavideD: I need the entity so that I can add it to the response

DavideD: I suppose I could return the `Response` directly

DavideD: I will change it to this:
```
       return Panache
                .withTransaction(() -> Fruit.<Fruit> findById(id)
                        .onItem().ifNotNull().transform(entity -> {
                            entity.name = fruit.name;
                            return Response.ok(entity).build();
                        }))
                .onItem().ifNull().continueWith(() -> Response.ok().status(NOT_FOUND).build());
```

DavideD: But I think I prefer the other approach as it keeps the update of the entity and the return of the response separated

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
DavideD: Ah ok, found it... `replaceWith`

DavideD: How come they have such a different name when they do the same thing?

DavideD: Ah ok, replace doesn't have the argument.

DavideD:
I've added a commit with the changes: https://github.com/quarkusio/quarkus-quickstarts/pull/809/commits/7d56eadc6201ac8443fa2949e1d89526b0b8558a

hibernate-reactive-panache-quickstart/src/main/java/org/acme/hibernate/orm/panache/FruitResource.java
```diff
@@ -0,0 +1,142 @@
...
+        if (fruit == null || fruit.name == null) {
+            throw new WebApplicationException("Fruit name was not set on request.", 422);
+        }
+
+        return Panache
+                .withTransaction(() -> Fruit.<Fruit> findById(id)
+                    .onItem().ifNotNull().transform(entity -> {
+                        entity.name = fruit.name;
+                        return Response.ok(entity).build();
+                    }))
```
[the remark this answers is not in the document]
DavideD: If I use `invoke` on the first line then I loose the entity.
Maybe I simply don't see it.

> We need to get away from using transform(), IMO.

But why? This to me seems a reasonable use case to me

DavideD:
I've updated the PR but there is still a `transform` that I don't think we can remove.
I'm happy to be proven wrong.

gavinking:
LGTM, thanks @DavideD!

FroMage:
@gavinking you need to approve the changes, if LGTM means that ;)

FroMage:
Well I can still merge it, though…

gavinking (approved):

PR #817, opened 2021-03-18
Fix s2i scripts

geoand:
These broke when we moved to fast-jar by default


cescoffier (approved):

geoand:
This was reported as bug from the Openshift folks BTW.

rsvoboda:
@geoand I think you could backport this fix into `main`, this way the "OCP embeded quickstart" would work once the fix gets merged

geoand:
@rsvoboda yeah, if we want the fix to be available now, it likely need to be merged into `main`

geoand:
@rsvoboda I pushed the change to `main` as well

PR #821, opened 2021-03-24
CI: Fix report condition

famod:
Just like here: https://github.com/quarkusio/quarkus/pull/15980


famod:
Will redo, wrong branch.

PR #847, opened 2021-04-26
Replace Debezium with Test Containers

cescoffier:


gsmet (approved):
Let's see what CI has to say.

cescoffier:
We may hit the Avro classloading issue, let's see...

gsmet:
Yeah but that's not an issue. I'll just check if the Kafka Streams test passes.

gsmet:
Kafka Streams is now OK AFAICS. Merging.

PR #849, opened 2021-04-28
Amazon SSM Quickstart

emattheis:
a new quickstart to go along with my Quarkus PR: https://github.com/quarkusio/quarkus/pull/16857

obviously it requires a local install of the extension introduced in that PR

most of this was copy/pasted from the existing quickstart for KMS, but I reworked the tests and the README


pom.xml
```diff
@@ -41,7 +42,6 @@
         <module>infinispan-client-quickstart</module>
         <module>jms-quickstart</module>
         <module>kafka-quickstart</module>
-        <module>kafka-avro-schema-quickstart</module>
```
gsmet, line 44: This doesn't look right.

pom.xml
```diff
@@ -60,7 +60,6 @@
         <module>opentracing-quickstart</module>
         <module>quartz-quickstart</module>
         <module>qute-quickstart</module>
-        <module>rest-client-reactive-quickstart</module>
```
gsmet, line 63: Same here.

gsmet (changes requested):
Thanks! I think you mistakenly removed two modules, could you have a look?

pom.xml
emattheis: not sure how I managed to do that 🤔 
must have been a bad merge
should be fixed now

emattheis:
@gsmet I restored those two modules (sorry about that!) and rebased - should be good to go

gsmet (approved):

gsmet:
Thanks! I started CI. There's a good chance it will be red due to other issues so don't panic :). I'll have a look once finished to check this particular project was built correctly.

gsmet:
And merged! Thanks!

PR #855, opened 2021-05-12
Parallelize native tests CI build

gsmet:
The main purpose of this is to reduce the time the build takes as the build started to take 6+ hours.

### DIFF
--- a/.github/generate-native-matrix.sh
+++ b/.github/generate-native-matrix.sh
@@ -1,0 +1,29 @@
+#! /bin/bash
+
+IFS=$'\n'
+modules=($(grep -LR --include=pom.xml '<packaging>pom</packaging>' */ | xargs dirname | sort))
+count=${#modules[@]}
+partition=$(expr $count / 4 + 1)
+
+json=$'{\n'
+json+=$'    "include": [\n'
+json+=$'    {\n'
+i=0
+partition_index=1
+for module in ${modules[@]}; do
+    if [ $i -gt $partition ]; then
+        json+=$'"\n    },\n    {\n'
+        i=0
+    fi
+    if [ $i -eq 0 ]; then
+        json+="        \"category\": \"Native Tests - $partition_index\","
+        json+=$'\n        "test-modules": "'
+        ((partition_index=partition_index+1))
+        i=0
+    fi
+    json+="$module,"
+    ((i=i+1))
+done
+json+=$'"\n    }\n    ]\n}\n'
+
+echo "$json"
--- a/.github/workflows/native-build-development.yml
+++ b/.github/workflows/native-build-development.yml
@@ -9,9 +9,27 @@ on:
   workflow_dispatch:
 
 jobs:
-  build_with_native:
+  generate-json-matrix:
+    name: Native Tests - Read JSON matrix
     runs-on: ubuntu-latest
     if: "github.repository == 'quarkusio/quarkus-quickstarts' || github.event_name == 'workflow_dispatch'"
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: generate
+        run: |
+          json=$(.github/generate-native-matrix.sh | tr -d '\n')
+          echo "::set-output name=matrix::${json}"
+  build-with-native:
+    name: ${{matrix.category}}
+    runs-on: ubuntu-latest
+    needs: [generate-json-matrix]
+    if: "github.repository == 'quarkusio/quarkus-quickstarts' || github.event_name == 'workflow_dispatch'"
+    strategy:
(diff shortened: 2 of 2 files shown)


.github/generate-native-matrix.sh
```diff
@@ -0,0 +1,29 @@
...
+partition_index=1
+for module in ${modules[@]}; do
+    if [ $i -gt $partition ]; then
+        json+=$'"\n    },\n    {\n'
+        i=0
+    fi
+    if [ $i -eq 0 ]; then
+        json+="        \"category\": \"Native Tests - $partition_index\","
+        json+=$'\n        "test-modules": "'
+        ((partition_index=partition_index+1))
```
famod, line 21: `partition_index+=1` should work as well but doesn't really matter

famod, line 21: `partition_index+=1` should work as well but doesn't really matter

.github/workflows/native-build-development.yml
```diff
@@ -9,9 +9,27 @@ on:
...
+    name: Native Tests - Read JSON matrix
     runs-on: ubuntu-latest
     if: "github.repository == 'quarkusio/quarkus-quickstarts' || github.event_name == 'workflow_dispatch'"
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: generate
+        run: |
+          json=$(./.github/generate-native-matrix.sh | tr -d '\n')
```
famod, line 22: I don't think the leading `./` is required here

.github/workflows/native-build-development.yml
```diff
@@ -90,11 +108,16 @@ jobs:
...
+    steps:
       - name: Report
-        if: "always() && github.repository == 'quarkusio/quarkus-quickstarts'"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_API_TOKEN }}
-          STATUS: ${{ job.status }}
+          STATUS: ${{ needs.build-with-native.result }}
         run: |
           echo "The report step got status: ${STATUS}"
           sudo apt-get update -o Dir::Etc::sourcelist="sources.list" \
```
famod, line 123: Could be switched to the less verbose way: https://github.com/quarkusio/quarkus/blob/main/.github/workflows/jdk-early-access-build.yml#L112-L119
But that's mostly offtopic for this PR.

.github/generate-native-matrix.sh
```diff
@@ -0,0 +1,29 @@
+#! /bin/bash
+
+IFS=$'\n'
+modules=($(find . -name pom.xml | sed -E 's@/pom.xml@@g' | sed -E 's@^\./@@g' | grep -vE '^\.$' | sort))
```
famod, line 4: This does the same:
```
find */ -name pom.xml -printf "%h\n" | sort
```
But both yield this:
```
kafka-streams-quickstart
kafka-streams-quickstart/aggregator
kafka-streams-quickstart/producer
```
(same for `funqy-quickstarts`)
I don't think it makes sense to execute those aggregators individually.
This should work:
```
grep -LR --include=pom.xml '<packaging>pom</packaging>' */ | xargs dirname | sort
```

famod (commented):
Great idea!

Btw, this reminds me to introduce GIB in this repo. WDYT?

.github/generate-native-matrix.sh
famod: PS: Using `-f` would allow keeping the `/pom.xml` suffix (no `dirname`, `-printf` or `sed` needed).

.github/generate-native-matrix.sh
gsmet: Tried it. Didn't work.

.github/workflows/native-build-development.yml
gsmet: this can be done in another PR.

.github/workflows/native-build-development.yml
gsmet: Yeah, removed.

.github/generate-native-matrix.sh
gsmet: Done.

.github/generate-native-matrix.sh
famod: Strange. So be it!

famod (approved):
LGTM now! There was a build failure previously but that did not look related to this change.

gsmet:
Yeah, I got the other issues fixed in other PRs.

famod:
@gsmet 
> Btw, this reminds me to introduce GIB in this repo. WDYT?

Any objections? If not I'll put it on my list...

gsmet:
I wouldn't say it's a top priority but I have nothing against it if you have the time to work on it :).

PR #861, opened 2021-05-18
Disable KafkaClientIT>KafkaClientTests.testBareClients in native

gsmet:
/cc @cescoffier


cescoffier:
Not very happy, but without a way to reproduce it (ran it 100 times in a row without issue), let's disable it.

PR #868, opened 2021-05-25
clean up HR example

gavinking:
Show better use of Mutiny APIs + reduce verbosity.


cescoffier (changes requested):
You need to target the "development" branch.

cescoffier:
The code looks good!

gavinking:
> The code looks good!

Yeah but the original version had a problem with error handling. In fixing that I got a bit annoyed that the tests don't seem to know about the `ErrorMapper`, which makes the code worse than it should be.

PR #869, opened 2021-05-25
minor changes to HR example

gavinking:
This had already been improved a lot on the development branch.

### DIFF
--- a/hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
+++ b/hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
@@ -20,8 +20,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
 
-import org.hibernate.reactive.mutiny.Mutiny;
-
+import org.hibernate.reactive.mutiny.Mutiny.Session;
 import org.jboss.logging.Logger;
 import org.jboss.resteasy.reactive.RestPath;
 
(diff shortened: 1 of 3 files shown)


gavinking:
The CI failures don't appear to be related to my changes.

gsmet:
@DavideD could you have look?

hibernate-reactive-quickstart/src/main/java/org/acme/hibernate/reactive/FruitMutinyResource.java
```diff
@@ -35,15 +34,12 @@
...
-    private static final Logger LOGGER = Logger.getLogger(FruitMutinyResource.class.getName());
 
     @Inject
-    Mutiny.Session session;
+    Session session;
 
     @GET
     public Uni<List<Fruit>> get() {
-        // In this case, it makes sense to return a Uni<List<Fruit>> because we return a reasonable amount of results
-        // Consider returning a Multi<Fruit> for result streams
```
DavideD, line 46: Why did you remove this comment? It's there because people are puzzled by the fact that we are not using a `Multi`

gavinking: Because the comment is confusing and misleading: "in this case"?? What is special about this case? We've been through this many times: `Multi` doesn't make sense with today's RDBMSs. The comment makes it seem like it would sometimes make sense, which is simply wrong.

DavideD: Several people have found it weird that we don't return a `Multi<Fruit>` in this case. The case in question is returning the result from a query or more precisely when to use `Multi<Fruit>` vs `Uni<List<Fruit>>`.

We had several discussion about this online but I think a comment here to mention this makes sense.
Now, the previous comment wasn't great but I would replace it with something better instead of removing it completely.

gavinking:
superseded by #1134

PR #871, opened 2021-05-26
create HR example using StatelessSession

gavinking:
A new example that shows off `StatelessSession`.


gsmet:
@DavideD could you have look?

DavideD:
Sure

DavideD:
Looks great but  maybe we could add a line to explain when one can use the `StatelessSession` instead of the regular one.
Otherwise I feel it might confuse people having two examples doing pretty much the same thing.

gavinking:
Why has this four-month old PR not been merged?

DavideD:
By the way, why is this a separate quickstart? Couldn't we add it in the one we already have?

PR #906, opened 2021-07-20
Instruction consistency - general polish

Sanne:
- consistently use `postgres:13.3` for the postgresql docker image (instead of many random different versions)
- remove `--ulimit memlock=-1:-1` and `--memory-swappiness=0` tuning from the postgresql instructions: problems on MacOS and really not necessary anymore
- use `./mvnw package` rather than `./mvnw install` when building the demos


cescoffier (approved):

Sanne:
thanks @cescoffier !

PR #925, opened 2021-08-24
Change OIDC quickstart to dev services

stuartwdouglas:


sberyozkin:
@stuartwdouglas Hi Stuart, this looks fine, thanks - would you like hibernate related updates going with this PR as well ?
Let me also try to encapsulate this token acquisition code into a light weight utility

hibernate-orm-quickstart/src/main/java/org/acme/hibernate/orm/FruitResource.java
```diff
@@ -43,7 +41,7 @@
 
     @GET
     @Path("{id}")
-    public Fruit getSingle(@PathParam Integer id) {
+    public Fruit getSingle(@PathParam("id") Integer id) {
```
gsmet, line 44: You don't have the equivalent in RESTEasy Reactive to avoid having the name?

hibernate-orm-quickstart/pom.xml
```diff
@@ -41,11 +41,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-resteasy</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-resteasy-jackson</artifactId>
+            <artifactId>quarkus-resteasy-reactive-jackson</artifactId>
```
gsmet, line 44: This is unrelated to the title of the PR... and while I think we should do it, we should probably do it all around the quickstarts if you and @geoand think it's a good idea to move the quickstarts to RESTEasy Reactive for 2.3.

I will be happy to do it as it will be a good opportunity to learn a bit more about it.

gsmet (changes requested):

hibernate-orm-quickstart/src/main/java/org/acme/hibernate/orm/FruitResource.java
geoand: You basically don't need any annotation at all :)

hibernate-orm-quickstart/pom.xml
geoand: Yeah, I think @cescoffier had a plan to do that for all of them

PR #931, opened 2021-09-03
Conscious Language Fixes

paulrobinson:
I only found these two things that could be fixed. All the other occurrences of problematic language were mostly in the book.txt (War and Peace) and in reference to Master Key (which I don't think is problematic).


paulrobinson:
I'm not sure why the build is failing, but it seems unrelated as these changes are just to markdown files.

gsmet (approved):

PR #939, opened 2021-09-17
Update Dockerfile.jvm

hurzelpurzel:
The Dockerfile didn't run with Kaniko on gitlab-ci. The separated chown did solve the problem.
I think Quarkus should be as compatible as possible and not only work with docker an builder.

**Check list**:

Your pull request:


cescoffier:
What's the error displayed by Kaniko?
The change looks good, but we would need to change the template used to generate the docker file.

hurzelpurzel:
error building image: error building stage: failed to get files used from context: failed to get fileinfo for /builds/playground108/logproducer/target/quarkus-app/lib: lstat /builds/playground108/logproducer/target/quarkus-app/lib: no such file or directory

As far as I understood, the --chown flag was introduced with docker-ce version 17.09 (see: https://stackoverflow.com/questions/44766665/how-do-i-docker-copy-as-non-root). It might be the case, that some builder like kaniko or moby does not support it.

cescoffier:
Looking at the issue, it seems to be a kaniko limitation. Hopefully, it will be fixed soon. 
Your approach works but introduces an additional layer, which, as the top layer, will be executed on every build. However, if there is no other choice, let's go for it.

cescoffier:
The PR does not target the right branch, please update to use the `development` branch.

hurzelpurzel:
switch to

PR #951, opened 2021-10-08
Map quarkus maven plugin version to the standard property

Sgitario:
This would allow QE to overwrite the version correctly.

**Check list**:

Your pull request:


Sgitario:
> In the CR4 `io.quarkus:quarkus-bom` and `io.quarkus:quarkus-maven-plugin` will share one version but `com.redhat.quarkus.platform:quarkus-bom` and `com.redhat.quarkus.platform:quarkus-maven-plugin` will share another version. But the versions of the quarkus-bom and the quarkus-maven-plugin are supposed to be always aligned.

This is exactly what we want to achieve: we want to use `com.redhat.quarkus.platform:quarkus-bom` to verify RHBQ bits and the problem is that some projects are using `quarkus.platform.version` for the quarkus maven plugin, so it will always use `io.quarkus:quarkus-maven-plugin:{quarkus.platform.version}`, so it's failing because `quarkus.platform.version` points to the `com.redhat.quarkus.platform:quarkus-bom` one.

Sgitario:
> This is what the config should look like
> 
> ```
>   <dependencyManagement>
>     <dependencies>
>       <dependency>
>         <groupId>${quarkus.platform.group-id}</groupId>
>         <artifactId>${quarkus.platform.artifact-id}</artifactId>
>         <version>${quarkus.platform.version}</version>
>         <type>pom</type>
>         <scope>import</scope>
>       </dependency>
>     </dependencies>
>   </dependencyManagement>
> ...
>   <build>
>     <plugins>
>       <plugin>
>         <groupId>${quarkus.platform.group-id}</groupId>
>         <artifactId>quarkus-maven-plugin</artifactId>
>         <version>${quarkus.platform.version}</version>
> ...
> ```

I agree that we should also replace the groupId to use `${quarkus.platform.group-id}`, but we should use `${quarkus-plugin.version}` as it's the standard parameter according to https://quarkus.io/guides/maven-tooling

aloubyansky:
> I agree that we should also replace the groupId to use `${quarkus.platform.group-id}`, but we should use `${quarkus-plugin.version}` as it's the standard parameter according to https://quarkus.io/guides/maven-tooling

I'll fix it right now. It won't be anymore.

aloubyansky:
Is there any other technical reason to re-introduce `quarkus-plugin.version` property?

Sgitario:
> Is there any other technical reason to re-introduce `quarkus-plugin.version` property?

Not really, but we have configured our jobs to replace the two properties (`quarkus-plugin.version` and `quarkus.platform.version`, separately).

So, @aloubyansky are you going to fix the five modules in a separate PR, or you're going to do only the replacing of the `groupId` property?

Sgitario:
> Our dev tools aren't including the `quarkus-plugin.version` property anymore in the POMs, so there is no point in mentioning it in the docs and the quickstarts should be adjusted accordingly too.

@aloubyansky PR updated to use:

```
<plugin>
                <groupId>${quarkus.platform.group-id}</groupId>
                <artifactId>quarkus-maven-plugin</artifactId>
                <version>${quarkus.platform.version}</version>
</plugin>

File: hibernate-orm-quickstart/pom.xml [hunk 1]
```diff
@@ -52,13 +52,6 @@
             <artifactId>quarkus-jdbc-postgresql</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-hibernate-orm-deployment</artifactId>
-            <scope>provided</scope>
-            <version>${quarkus-plugin.version}</version>
-        </dependency>
```
Sgitario, line 60: This dependency was not necessary: It worked fine without it.

aloubyansky (approved):

Sgitario:
> @Sgitario we will need the same for 2.2 branch ;)

Sure, I will backport it once this PR is merged. 
@aloubyansky the CI is passing, I think this PR is good to be merged now.

maxandersen:
> Our dev tools aren't including the `quarkus-plugin.version` property anymore in the POMs, so there is no point in mentioning it in the docs and the quickstarts should be adjusted accordingly too.

Is it not still relevant for Gradle ?

PR #973, opened 2021-10-28
Update keycloak authorization tests to use DevServices and make the realm file accessible as a classpath resource in devmode

### LINKED ISSUE #972: security-keycloak-authorization-quickstart Guide and application.properties should be updated to https
in [ https://github.com/quarkusio/quarkus-quickstarts/blob/main/security-keycloak-authorization-quickstart/src/main/resources/application.properties ]

we have this line

quarkus.oidc.auth-server-url=https://localhost:8543/auth/realms/quarkus

but in [ https://github.com/quarkusio/quarkus-quickstarts/blob/main/security-keycloak-authorization-quickstart/README.md ]

we use the port : 8180    
http://localhost:8180/auth/realms/quarkus/protocol/openid-connect/token

I checked the git history and I did a rollback and the sample in running now.

quarkus.oidc.auth-server-url=http://localhost:8180/auth/realms/quarkus

---------
from Sergey

I recall now it was done to test keycloak authorization can work over HTTPS with the TLS verification disabled, there is a problem with using the Keycloak Admin Client with HTTPS so I could not test the Quarkus/main.
Does it work for you if, with the https: address, if you also add `quarkus.oidc.tls.verification=none` (as in the quickstart properties) ?
IMHO it would make sense to update the guide itself to show `https:` can be used...

sberyozkin:
Fixes #972 

This PR provides a few improvements to OIDC quickstarts related to how they run with Dev Services, following the conversation with Daniel Oh

- Currently directly accessing `quarkus-realm.json` as a classpath resource does not work in `quarkus:dev` so with a hint from Alexey I got it working for OIDC demos
- When one does `quarkus:dev` and goes to Keycloak Dev UI and tries to get the access tokens via the code flow Keycloak does not redirect back since the imported realm does not allow it so I had to ask Daniel to change the real to allow a `*` redirect and he did it twice which was not cool :-)
- Updated keycloak-authorization to test with Dev Services - same way it is done for OIDC and OIDC web-app demos


janpfeil (commented):
adoc syntax was used in markdown file, renders ok but looks broken

PR #980, opened 2021-11-05
Standardize the examples

survivant:
I updated the sample to have the same structure.  It's easier to compare between them.

### DIFF
--- a/hibernate-orm-jakarta-data-quickstart/pom.xml
+++ b/hibernate-orm-jakarta-data-quickstart/pom.xml
@@ -78,6 +78,8 @@
                 <configuration>
                     <!-- the parameters=true option is critical so that RESTEasy works fine -->
                     <parameters>true</parameters>
+                    <!-- for the annotation processor transitive dependencies to be managed by Quarkus, we need this setting -->
+                    <annotationProcessorPathsUseDepMgmt>true</annotationProcessorPathsUseDepMgmt>
                     <!-- the hibernate annotation processor provides both static metamodel and Jakarta Data capabilities: -->
                     <annotationProcessorPaths>
                         <path>


survivant:
@sberyozkin here the new PR for the development branch.  CC @gsmet

security-keycloak-authorization-quickstart/src/main/java/org/acme/security/keycloak/authorization/AdminResource.java
```diff
@@ -1,3 +1,18 @@
+/**
```
sberyozkin, line 1: @survivant I'm not sure the license headers have to be added to the main source files, if they have to then it is likelyly have to be `2019-2021` years, but please wait for @gsmet to confirm

security-openid-connect-quickstart/pom.xml
```diff
@@ -32,6 +32,10 @@
     </dependencyManagement>
 
     <dependencies>
+        <dependency>
```
sberyozkin, line 35: @survivant Why is this module added here ? There is a dedicated `keycloak-authorization-quickstart`

security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/BearerTokenAuthenticationTest.java
```diff
@@ -10,29 +10,42 @@
...
-    public void testAdminAccess() {
+    public void testUserAccess() {
+        RestAssured.given().auth().oauth2(getAccessToken("alice"))
+                .when().get("/api/users/me")
+                .then()
+                .statusCode(200);
+
         RestAssured.given().auth().oauth2(getAccessToken("admin"))
-                .when().get("/api/admin")
+                .when().get("/api/users/me")
```
sberyozkin, line 25: @survivant Hmm but this drops a check that `alice` (who only has a user role) can not access `admin`-level resources

security-openid-connect-quickstart/src/test/java/org/acme/security/openid/connect/BearerTokenAuthenticationTest.java
```diff
@@ -10,29 +10,42 @@
...
+                .statusCode(200);
     }
 
     @Test
-    public void testUserAccess() {
-
+    public void testAdminAccess() {
         RestAssured.given().auth().oauth2(getAccessToken("alice"))
-                .when().get("/api/users/me")
+                .when().get("/api/admin")
```
sberyozkin, line 38: Similarly for `admin`

security-keycloak-authorization-quickstart/src/main/java/org/acme/security/keycloak/authorization/AdminResource.java
```diff
@@ -1,3 +1,18 @@
...
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
```
gsmet, line 15: We don't add license headers in Quarkus. They shouldn't be added.

survivant:
@sberyozkin the goal was to get the samples : security-openid-connect-quickstart  and security-keycloak-authorization-quickstart  almost similar except one use the token and the other roles.

security-keycloak-authorization-quickstart/src/main/java/org/acme/security/keycloak/authorization/AdminResource.java
survivant: good, I can do a cleanup in the other samples too.

survivant: there are 14 others files with licence 

image

security-openid-connect-quickstart/pom.xml
survivant: @sberyozkin maybe it's not necessary.  I though it was used by KeycloakTestClient.  So the sample : **security-keycloak-authorization-quickstart** need that module to check in Keycloak if a URL is authorized for the user logged.  

and the sample : **security-openid-connect-quickstart**   will received a token with the roles in it and use that to validate if the user is authorized with the annotation @RolesAllowed so the dependency to keycloak autorisation is not needed ?

did I understand correctly the difference between the two samples ?

security-keycloak-authorization-quickstart/src/main/java/org/acme/security/keycloak/authorization/AdminResource.java
survivant: removed

PR #986, opened 2021-11-15
Fix S3 quickstart

gsmet:
For whatever reason, there's an additional character and I switched to
containsString().


amazon-s3-quickstart/src/test/java/org/acme/s3/S3ResourcesTest.java
```diff
@@ -58,7 +59,7 @@ void testResource(final String testedResource) {
                 .when().get("/{resource}/download/{objectKey}")
                 .then()
                 .statusCode(200)
-                .body(equalTo(fruit))
+                .body(containsString(fruit))
```
cescoffier, line 62: shouldn't it be startsWith?

gsmet: Well, TBH, I don't care :).

PR #994, opened 2021-11-17
Update OIDC web-app quickstart to use the client secret, remove duplicate realm files

sberyozkin:
It relates to https://github.com/quarkusio/quarkus/issues/21500. The real problem is that the OIDC `web-app` applications should not use public clients since such applications are OIDC `confidential clients`. The user tried to add a secret to the configuration and mistyped - so this PR does the following
1) updates the web-app guide's realm and `application.properties` to use a client with a secret
2) not really related to the issue - but also added a `permit` policy for `/q/dev` - without it, with `quarkus:dev`, I had to login to Keycloak twice (once when going to `q/dev` and once - when clicking on `Keycloak` in the OIDC card :-) )
3) Took me awhile to figure out why, by updating the web-app guide's realm (`config/quarkus-realm.json`) I could not get the tests failing - proved that each of the OIDC guides had a duplicate in `test/resources` - it is no longer needed since `config/realm.json` is copied to `classes` during the build now

I'll follow up shortly with a minor update to the OIDC web-app doc as well


gastaldi (approved):

fedinskiy:
@sberyozkinm sorry for ressurecting this discussion, but  what was the purpose if this[1] change?

It leads to the session being limited to two seconds, which practically means that the user needs to re-login every time she accesses protected url, which sort of defeats the purpose of being logged in :) .
This file is used in tutorial[2], so we are confusing everyone who starts working with OIDC.

[1] https://github.com/quarkusio/quarkus-quickstarts/pull/994/files#diff-60e41dbf53b08da430618f1b24ee64fbdc42a9f22c9817ecb5a0b8aa556facf8R10
[2] https://quarkus.io/version/main/guides/security-oidc-code-flow-authentication-tutorial

PR #1001, opened 2021-12-06
Update OIDC web-app quickstart to test the realm file loaded from the file system

sberyozkin:
Related to #21975.

The test uses Dev Services for Keycloak and now the realm file is loaded from the file system.

`security-openid-connect-quickstart` and `security-keycloak-authorization-quickstart` will continue testing with realm file loaded from the classpath


sberyozkin:
Closing it as it will be tested now directly in Quarkus

PR #1017, opened 2021-12-15
Update Maven wrapper to 3.8.4

gsmet:
Or install it where it was missing.


famod (approved):
LGTM

I haven't checked every single file. I wasn't sure about those funqy quickstarts but I guess it makes sense to run those archetypes via mvnw as well.

Btw, I just tested the "official" wrapper: https://lists.apache.org/thread/bkgdbwvf34kl67dbnhvk2l1cdz5s1gs9
```
mvn org.apache.maven.plugins:maven-wrapper-plugin:3.1.0-SNAPSHOT:wrapper -Papache-snapshots
```
which seemed to work, but we can always do another iteration.

gsmet:
Yeah changing the wrapper mandates a discussion IMHO as we will have to do it everywhere (repositories, project template creation...)

PR #1045, opened 2022-02-02
Use quay.io/bitnami/postgresql to lower pressure on DockerHub limits

rsvoboda:
Use quay.io/bitnami/postgresql to lower pressure on DockerHub limits

Your pull request:


gsmet (changes requested):
Did we ever have issues in the quickstarts?

Because I'm not a big fan of using non-official images.

rsvoboda:
The official ones are used in dev ervices, this change affects just 2 modules.

We are hitting pull limits for DockerHub when we run QS, we have clean machines without any image.
The issue wouldn't be elimitad even with these changes in place, so maybe we can ignore this PR

PR #1067, opened 2022-02-23
Parametrize Quarkus Platform and plugin version

pjgg:
This PR fix some cosmetic issues in some poms, plus parametrize Quarkus Platform and plugin version. This could be useful in order to overwrite the default versions by system properties.

Additionally, kogito group ID is also parametrized

### DIFF
--- a/amazon-dynamodb-quickstart/pom.xml
+++ b/amazon-dynamodb-quickstart/pom.xml
@@ -12,12 +12,14 @@
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
         <quarkus.platform.version>999-SNAPSHOT</quarkus.platform.version>
-        <surefire-plugin.version>3.0.0-M5</surefire-plugin.version>
+        <quarkus-plugin.version>${quarkus.platform.version}</quarkus-plugin.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <quarkus-amazon-services.version>1.0.4</quarkus-amazon-services.version>
         <awssdk.testcontainers.version>1.12.57</awssdk.testcontainers.version>
+        <surefire-plugin.version>3.0.0-M5</surefire-plugin.version>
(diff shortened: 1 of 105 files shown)


amazon-dynamodb-quickstart/pom.xml
```diff
@@ -12,12 +12,14 @@
...
         <quarkus.platform.version>999-SNAPSHOT</quarkus.platform.version>
-        <surefire-plugin.version>3.0.0-M5</surefire-plugin.version>
+        <quarkus-plugin.version>${quarkus.platform.version}</quarkus-plugin.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <quarkus-amazon-services.version>1.0.4</quarkus-amazon-services.version>
         <awssdk.testcontainers.version>1.12.57</awssdk.testcontainers.version>
+        <surefire-plugin.version>3.0.0-M5</surefire-plugin.version>
+        <failsafe-plugin.version>${surefire-plugin.version}</failsafe-plugin.version>
```
gsmet, line 22: How similar is it to what we have when we create a project with the tooling? Because I think we should try to align things with that as much as we can.

pjgg: The values of all of these params are the same, but by this implementation, we are able to overwrite some parameters by system properties. 

for example,

`mvn clean install -Dquarkus.platform.group-id=com.redhat.quarkus.platform -Dquarkus.platform.version=2.2.5.Final-redhat-00007  -Dquarkus.platform.artifact-id=quarkus-bom`

pjgg:
After talking with Guillaume, we conclude that we are trying to keep the poms files, the most similar to the ones that are generated by [Quarkus client](https://quarkus.io/guides/cli-tooling). 

That means, that `quarkus-maven-plugin` version and `quarkus.platform.version` should have the same version, so doesn't make sense to parametrize `quarkus-maven-plugin` version. Also `surefire-plugin.version` and `failsafe-plugin.version` has the same version value. 

From now we are gonna close this PR, and I will verify that `quarkus-maven-plugin` version is not hardcoded in any module as well as `quarkus.platform.artifact-id`.

rsvoboda:
@pjgg but https://github.com/quarkusio/quarkus-quickstarts/pull/1067/commits/748ad0a4e024ce90b9e21cbcaacd949304bcb523 is relevant, there should be dedicated PR for this change

PR #1072, opened 2022-03-03
Migrate the context propagation quickstart to RESTEasy Reactive

cescoffier:
@gsmet FYI, this is what is required to migrate the context propagation QS to resteasy reactive.

Here ate the changes:

1. the `pom.xml` - nothing surprising
2. the emitter part is more an improvement (not required, but better - more idiomatic)
3. the SSE part is where the main change is. We have to use `.emitOn` as it is called on the IO thread that you are not allowed to block. Thanks to CP, you can use `.emitOn` and the transaction will be propagated.

### DIFF
--- a/context-propagation-quickstart/pom.xml
+++ b/context-propagation-quickstart/pom.xml
@@ -34,11 +34,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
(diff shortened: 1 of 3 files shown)


gsmet:
Closing this one as I will incorporate it into the unified PR.

PR #1108, opened 2022-05-08
Add ExceptionMapper to security-openid-connect-client-quickstart

sberyozkin:


security-openid-connect-client-quickstart/src/main/resources/application.properties
```diff
@@ -11,6 +11,10 @@ quarkus.oidc-client.grant.type=password
 quarkus.oidc-client.grant-options.password.username=alice
 quarkus.oidc-client.grant-options.password.password=alice
 
-org.acme.security.openid.connect.client.ProtectedResourceOidcClientFilter/mp-rest/url=http://localhost:8081/protected
-org.acme.security.openid.connect.client.ProtectedResourceTokenPropagationFilter/mp-rest/url=http://localhost:8081/protected
+%prod.port=8080
```
sberyozkin, line 14: @gsmet Guillaume, is it a good way to handle client ports such that it works in dev, test and prod modes ? I've verified, it works, but may be there is a neater way

Ladicek: You should be able to just use `${quarkus.http.port}` I think?

sberyozkin: @Ladicek No, it does not work, tests fail as the test endpoints are on `8081` but `quarkus.http.port` is `8080` by default.
I guess I can set  `quarkus.http.port` to `8081` in pom.xml but I'm not sure if it will be a simpler approach ?

Ladicek: Sorry, I thought `quarkus.http.port` is 8081 in the `test` profile. I swear I saw something like this before, but not sure _exactly_...

sberyozkin: @Ladicek No problems, thanks for the idea, it makes sense to have it defaulting to `8081`, may be it was done indirectly by assigning it to `quarkus.http.test-port` in some demo configuration

gsmet (approved):
Yeah, it's not pretty, we probably have to improve on this but it will do for now.

PR #1117, opened 2022-05-24
Fix security-openid-connect-client-quickstart in native

geoand:
Fixes: https://github.com/quarkusio/quarkus/issues/25731

### DIFF
--- a/security-openid-connect-client-quickstart/src/main/resources/application.properties
+++ b/security-openid-connect-client-quickstart/src/main/resources/application.properties
@@ -11,10 +11,6 @@ quarkus.oidc-client.grant.type=password
 quarkus.oidc-client.grant-options.password.username=alice
 quarkus.oidc-client.grant-options.password.password=alice
 
-%prod.port=8080
-%dev.port=8080
-%test.port=8081
-
-org.acme.security.openid.connect.client.ProtectedResourceOidcClientFilter/mp-rest/url=http://localhost:${port}/protected
-org.acme.security.openid.connect.client.ProtectedResourceTokenPropagationFilter/mp-rest/url=http://localhost:${port}/protected
+org.acme.security.openid.connect.client.ProtectedResourceOidcClientFilter/mp-rest/url=${test.url}/protected
+org.acme.security.openid.connect.client.ProtectedResourceTokenPropagationFilter/mp-rest/url=${test.url}/protected
 


gsmet (approved):

security-openid-connect-client-quickstart/src/main/resources/application.properties
```diff
@@ -11,10 +11,6 @@ quarkus.oidc-client.grant.type=password
...
 quarkus.oidc-client.grant-options.password.password=alice
 
-%prod.port=8080
-%dev.port=8080
-%test.port=8081
-
-org.acme.security.openid.connect.client.ProtectedResourceOidcClientFilter/mp-rest/url=http://localhost:${port}/protected
-org.acme.security.openid.connect.client.ProtectedResourceTokenPropagationFilter/mp-rest/url=http://localhost:${port}/protected
+org.acme.security.openid.connect.client.ProtectedResourceOidcClientFilter/mp-rest/url=${test.url}/protected
+org.acme.security.openid.connect.client.ProtectedResourceTokenPropagationFilter/mp-rest/url=${test.url}/protected
```
gsmet, line 15: Hmmm wait. There is an issue: Sergey wanted the app to work if you run it. And with this, it's broken unfortunately.

geoand: Hm.... Good point. We'll likely have to have a `%prod.org.acme.security.openid.connect.client.ProtectedResourceOidcClientFilter/mp-rest/url`. @sberyozkin mind taking care of it?

gsmet: Well, won't the problem be the same? My guess is that we cannot have both the native tests and the app to be run successfully :/

geoand: We can use a different run profile for the native test.

sberyozkin: @geoand Thanks for a fix, but I'm sorry, I don't think it makes things better, it fails straight away in a dev mode:
```
One or more configuration errors have prevented the application from starting. The errors are:
  - SRCFG00011: Could not expand value test.url in property org.acme.security.openid.connect.client.ProtectedResourceOidcClientFilter/mp-rest/url
  - SRCFG00011: Could not expand value test.url in property org.acme.security.openid.connect.client.ProtectedResourceTokenPropagationFilter/mp-rest/url
```

I guess it won't work when runining in JVM or native mode either, I believe this is what Guillaume said.

The original issue was that a native test was failing due to a port mismatch, with the endpoint picking up a prod profile value. IMHO a better solution is to ensure that in native mode the test port is also allocated to the endpoint.

I think we should revert and review the options again

geoand: We can use a specific profile for the native test (by using `quarkus.test.native-image-profile`).

sberyozkin: @geoand cool, that property makes a difference, let me open a new PR

geoand: Excellent!

PR #1118, opened 2022-05-24
Set quarkus.test.native-image-profile to 'test' in the OidcClient quickstart

sberyozkin:
I've verified this configuration works with tests (JVM, native), in devmode (and tested the token propagation by invoking the endpoint from DevServices for Keycloak console), when running in JVM and native modes (and starting Keycloak and using curl to test)


sberyozkin:
Thanks Georgios for the new property idea

geoand:
👍🏼

PR #1137, opened 2022-07-02
Update surefire/failsafe to 3.0.0-M7

famod:


famod:
I'll go ahead and merge this because the main repo is already using this version, including codestarts etc.

PR #1138, opened 2022-07-02
Update to Maven 3.8.6 (via Maven Wrapper 3.1.1)

famod:


albers:
Is there a specific reason for deleting the `MavenWrapperDownloader` and adding _maven-wrapper.jar_ instead?

IMHO binaries should be kept out of version control, and dropping an existing solution for downloading required binaries is no improvement.

The maven-generated quickstart uses the `MavenWrapperDownloader`, so for consistency, I'd recommend to keep this solution here.

famod:
> Is there a specific reason for deleting the `MavenWrapperDownloader` and adding _maven-wrapper.jar_ instead?

The downloader was removed by `mvn org.apache.maven.plugins:maven-wrapper-plugin:3.1.1:wrapper -Dmaven=3.8.6` (see also [plugin description](https://maven.apache.org/wrapper/maven-wrapper-plugin/wrapper-mojo.html)). I didn't do it myself manually. I haven't checked why it removes it in detail, will try to find out more.
The same was done for the main repo, btw.

Please note that I did _not add_ `maven-wrapper.jar`, it was there before, e.g.: https://github.com/quarkusio/quarkus-quickstarts/tree/2.10.1.Final/.mvn/wrapper
I was only updated.
 
> IMHO binaries should be kept out of version control, and dropping an existing solution for downloading required binaries is no improvement.

I agree in general, but that decision had been made before this PR, conciously or not I don't know/remember but I think it was on purpose.
I think small wrappers like this one are a bit of a gray area though.

> The maven-generated quickstart uses the `MavenWrapperDownloader`, so for consistency, I'd recommend to keep this solution here.

That's a good catch, but to me it doesn't mean automatically that the downloader needs to stay here. Maybe the generated quickstart has to be adjusted instead.

/cc @gsmet @aloubyansky @gastaldi

famod:
> I haven't checked why it removes it in detail, will try to find out more.

It's part of https://github.com/apache/maven-wrapper/pull/13 so I suppose it's on purpose.
It kinda makes sense to me because why would you want to keep the Downloader around when the wrapper has already been downloaded? (again, having committed the jar to git was already the case before this PR)

albers:
Thanks @famod for clarifying. I misread the diff in github (showing only added bytes for the binary), and I did not know about the _maven-wrapper-plugin_. Nice tool.

The current state of the repo regarding the maven wrapper is a mix between the _bin_ and _source_ [installation types](https://maven.apache.org/wrapper/maven-wrapper-plugin/usage.html#apache-maven-wrapper-distribution-types) offered by the _maven-wrapper-plugin_: both _MavenWrapperDownloader.java_ and _maven-wrapper.jar_ are present.

I think it's best to use the plugin for restoring a consistent state. We have two options here:

```shell
mvn org.apache.maven.plugins:maven-wrapper-plugin:3.1.1:wrapper -Dmaven=3.8.6 -Dtype=source
```
will update _MavenWrapperDownloader.java_ and remove _maven-wrapper.jar_ (which I prefer),

```shell
mvn org.apache.maven.plugins:maven-wrapper-plugin:3.1.1:wrapper -Dmaven=3.8.6
```
will update _maven-wrapper.jar_ and remove _MavenWrapperDownloader.java_ (the _bin_ type as used in this PR).

I think it boils down to the question how reliable the downloader is.

famod:
> I think it's best to use the plugin for restoring a consistent state.

Well, this PR does, but not in your preferred way. 😉
It boils down to whether we want to have the jar in git or not. Not having it in git increases the likelyhood of download issues, so one could argue that this defeats the point of a "quick start".

> I think it boils down to the question how reliable the downloader is.

FYI https://maven.apache.org/wrapper/#usage-without-binary-jar:

> The download is attempted via curl, wget and, as last resort, by compiling the ./mvn/wrapper/MavenWrapperDownloader.java file and executing the resulting class.

>  Do we have problem reports about the codestarts (that use the downloader and do not contain the wrapper)?

I don't know. Maybe @ia3andy knows.

gastaldi (approved):
I think this is fine

ia3andy:
We always add the wrapper in generated apps.

albers:
There seems to be a consensus to keep the _maven-wrapper.jar_ in the repo.
Under these conditions, I'm fine with this PR because it restores a consistent state.

> Maybe the generated quickstart has to be adjusted instead.

I do not support this idea because it will make people accidentally checking in the _maven-wrapper.jar_ in their initial commit because they do not expect binaries in a starter project.

IMHO the default should be a generated project without the binary, requiring people who prefer to have the wrapper in the codebase to perform an additional step.

PR #1140, opened 2022-07-11
Improve names for Rest Clients in security-openid-connect-client-quickstart

albers:
I and a collegue of mine found in unnecessarily hard to understand the security-openid-connect-client-quickstart.

This was caused by confusing class and field names.

### About the example

The `FrontendResource` uses two MP REST clients in order to access the `ProtectedResource`.
The  only relevant difference between the two REST clients is that they use different `ClientRequestFilter`s: one uses the _OpenID Connect Client Reactive Filter_ and the other uses the _OpenID Connect Token Propagation Reactive Filter_.

### What confused me

The two REST client interfaces used in `FrontendResource` are called `ProtectedResourceOidcClientFilter` and `ProtectedResourceTokenPropagationFilter`. 
In the context of an example on how to use two quarkus-supplied `ClientRequestFilter`s, these names cause confusion.
It requires mental mapping to understand that the two `*Filter` classes are not filters but REST clients (that use Filters).

### Suggestion

This PR renames the classes, giving them names that communicate the intent of the example more clearly.

### Follow-up

I will create a follow-up PR in the Quarkus repo to update the [guide](https://quarkus.io/guides/security-openid-connect-client) accordingly once this PR is merged.

### DIFF
--- a/security-openid-connect-client-quickstart/src/main/java/org/acme/security/openid/connect/client/FrontendResource.java
+++ b/security-openid-connect-client-quickstart/src/main/java/org/acme/security/openid/connect/client/FrontendResource.java
@@ -13,37 +13,37 @@ import io.smallrye.mutiny.Uni;
 public class FrontendResource {
     @Inject
     @RestClient
-    ProtectedResourceOidcClientFilter protectedResourceOidcClientFilter;
+    RestClientWithOidcClientFilter restClientWithOidcClientFilter;
 
     @Inject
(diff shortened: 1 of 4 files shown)


rsvoboda (approved):

gsmet:
Thanks!

albers:
Corresponding PR for the guide changes: https://github.com/quarkusio/quarkus/pull/28382

PR #1159, opened 2022-08-22
Get OIDC quickstarts working with 19.0.1

sberyozkin:


sberyozkin:
Hey @pedroigor Thanks, interesting point, let me play a bit with it with devservices

gsmet (changes requested):
Let's make sure we only merge this after I push the 2.12.0 quickstarts as the Keycloak 19 upgrade is for 2.13.

gsmet (approved):
I released the 2.12 quickstarts so we can merge this!

PR #1179, opened 2022-09-18
Close the Kafka Consumer when terminating in the Bare Kafka quickstart

scholzj:
The Kafka Consumer in the `bare-kafka-quickstart` is polling Kafka in a while loop based on `!done`. So it runs as long as the `done` is set to `false`. When terminating, the endpoint sets the `hold`, but it sets it to `false` again, so the Kafka Consumer will keep running. I assume this is just a typo and it should be set to `false` to close the consumer as well and not just the producer and admin clients..

**Check list**:

Your pull request:


cescoffier (approved):
Thanks!

cescoffier:
@scholzj Thanks! Yes, definitely a typo.

gsmet (approved):

PR #1199, opened 2022-11-02
Use Multi over raw RS Publisher

jponge:
- This anticipates the future switch to JDK Flow APIs.
- Once this happens then code depending on legacy RS APIs will need Mutiny Zero adapters (e.g., ReactiveFruitResource / Neo4J).


gsmet:
We probably need to also check the associated guides.

I will let @cescoffier check if that's all right and make sure we don't actually want to show the `Publisher` approach.

cescoffier:
I agree with the approach. We should hide Publisher as much as we can. It's an interop type, not an user-facing type.

jponge:
Noted for guides, I'll make an issue

PR #1204, opened 2022-11-04
update quarkus.hibernate-orm.database.generation to drop-and-create i…

soul2zimate:
…n hibernate-orm-quickstart

Avoid `org.postgresql.util.PSQLException: ERROR: relation "known_fruits" already exists`


yrodiere:
Sorry but I disagree, `drop-and-create` is not an appropriate value for a production environment.

I'll close this PR, feel free to answer if you think we should discuss this further.

PR #1229, opened 2023-01-17
Missing protocol in the mocked kubernetes environment for the Stork Kubernetes quickstart

cescoffier:
\CC @gsmet


kafka-quickstart/processor/pom.xml
```diff
@@ -34,6 +34,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-smallrye-reactive-messaging-kafka</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-health</artifactId>
+        </dependency>
```
gsmet, line 40: I think this change is unrelated? I don't mind adding it if it's not a mistake.

aureamunoz: Yes, I have the same question

cescoffier: hum.... sorry, I didn't stage that but the github cli did. Let me remove it.

PR #1252, opened 2023-03-22
hibernate-search-orm-elasticsearch - reset Author_SEQ to start with 3

rsvoboda:
hibernate-search-orm-elasticsearch - reset Author_SEQ to start with 3

Import script inserts authors with id 1 and 2, the sequence should start with 3 then.

This got reveleaed with the latest changes in Quarkus main, I thinks it's https://github.com/quarkusio/quarkus/pull/31947 done by @yrodiere, but didn't bisect Quarkus main to be sure.


yrodiere (approved):
Thanks.

It got me worried that my PR triggered this failure, so I investigated a bit. And it turns out the error was there all along, since a9ed414f612e6ca315672662bf5742b2d7df267b. It's just that because we were using the `pooled` optimizer (the default in ORM), ORM was using a pool from something like `-47` to `2`, starting with `-47`, so we wouldn't have seen the problem until we created more than 50 entities.

To be perfectly correct, a9ed414f612e6ca315672662bf5742b2d7df267b should have reset the sequence to something like `53`.
Now that we use the `pooled-lo` optimizer by default, though, resetting to `3` is the right choice.

PR #1257, opened 2023-04-04
Quarkus 3

gsmet:


gsmet:
@scrocquesel would you have an idea of what's going on here with the S3 test failing?

scrocquesel:
> @scrocquesel would you have an idea of what's going on here with the S3 test failing?

@gsmet at first glance, it seems an issue with localstack. I can't reproduce it with the current latest tag (updated 4 hours ago), but with `quarkus.aws.devservices.localstack.image-name=localstack/localstack:2.0.0` I have the same error.

Could you retry the test run ?

PR #1317, opened 2023-08-17
Artifacts need to be readable/executable as part of the root group in…

Jaland:
… order to be used on an openshift container by default.

**Check list**:

Your pull request:

- [x ] targets the `development` branch

Currently, if you try to use this docker file on Openshift you get a permissions error when trying to run `run-java.sh`. This is because the root group does not have access to execute that file. I don't think giving the root group access hurts the usage on other platforms but it will reduce the work for people trying to deploy onto Openshift.


gsmet:
Thanks for this PR. I think we will have to also apply that to the project generation in the core repository.

@maxandersen do you know someone that would be able to be authoritative on these? I have seen some back and forth 
regarding these permissions and it would be nice to have them settled once and for all with the current state of art in mind.

maxandersen:
Hmm. What version of openshift is it not working by default?

We already test and use this on multiple openshift versions so would want to know more before changing it (again) :)

PR #1322, opened 2023-08-24
Adapt QuickStarts for remote Docker agent on a same network

michalvavrik:
I need to run Quarkus Quickstarts on a remote Docker agent (in order to test Quarkus on Windows). That means everywhere that is hardcoded `localhost` doesn't work for me. This PR enables overriding host. This is important especially on Windows when using Docker Desktop alternatives as some of them use different Docker host. Fixed ports are no issue for me (I use NAT) and it would take much harder work to make it work for remote Docker agents that can be exposed on random computer port (f.e. with TestContainers Cloud).

Changes should be self-explanatory except for `mqtt-quickstart`. I can't set `mp.messaging.outgoing.topic-price.host` and `mp.messaging.incoming.prices.host` directly for it breaks `pulsar-quickstart-producer`. Allowing override host that is already done elsewhere (f.e. `hibernate-orm-multi-tenancy-quickstart` with `DB_HOST_TENANT_BASE` and `DB_HOST_TENANT_MYCOMPANY`) makes sense to me.

**Check list**:

Your pull request:

- [x] has integration/native tests (`mvn clean verify -Pnative`)
- [x] makes sure the associated guide must not be updated

### DIFF
--- a/mqtt-quickstart/src/main/resources/application.properties
+++ b/mqtt-quickstart/src/main/resources/application.properties
@@ -3,13 +3,13 @@
 # Configure the MQTT sink (we write to it)
 mp.messaging.outgoing.topic-price.connector=smallrye-mqtt
 mp.messaging.outgoing.topic-price.topic=prices
-mp.messaging.outgoing.topic-price.host=localhost
+mp.messaging.outgoing.topic-price.host=${MQTT_HOST:localhost}
 mp.messaging.outgoing.topic-price.port=1883
 mp.messaging.outgoing.topic-price.auto-generated-client-id=true
 
 # Configure the MQTT source (we read from it)
 mp.messaging.incoming.prices.connector=smallrye-mqtt
 mp.messaging.incoming.prices.topic=prices
-mp.messaging.incoming.prices.host=localhost
+mp.messaging.incoming.prices.host=${MQTT_HOST:localhost}
 mp.messaging.incoming.prices.port=1883
 mp.messaging.incoming.prices.auto-generated-client-id=true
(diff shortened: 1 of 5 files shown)


rsvoboda (approved):
Thank you for the contribution!

rsvoboda:
@gsmet we need this in 3.2.5.Final tag. There is no 3.2 branch yet.

Can you cherry-pick this commit before you do 3.2.5.Final tag?

gsmet:
Well, just to be clear: this is at the very limit of what I’m ready to accept in the QuickStarts: they are supposed to be kept simple and in line with the documentation.  

So please don’t complexity them. If you need to do it, you are probably using them for something they weren’t designed for.

michalvavrik:
> Well, just to be clear: this is at the very limit of what I’m ready to accept in the QuickStarts: they are supposed to be kept simple and in line with the documentation.
> 
> So please don’t complexity them. If you need to do it, you are probably using them for something they weren’t designed for.

I understand the complexity part and agree @gsmet , but I think you are factually wrong for this changes are proportional to what is in the quickstarts present. Please let's revise facts:

- simplify `security-jdbc-quickstart` and I dare say this is a proper way to start PG container
- changes in `rabbitmq-quickstart` are doing injection property instead of hardcoding stuff. Using `@ConfigProperty` is spread over all quickstarts and for example `amqp-quickstart` has literally absolutely identical lines except it use `amqp-host` and `amqp-port` instead of `rabbitmq-host`. TBH that's where I got it from
- changes in `mqtt-quickstart` are doing the same thing as is done in `hibernate-orm-multi-tenancy-quickstart`, which is stated in the PR description

So not only this PR brings improvement, but also as I made clear - not everyone is using Linux or Docker Desktop (it's paid!) and that's when it becomes important to be able to not hardcode localhost.

gsmet:
@michalvavrik I might have been unclear but when I said the `very limit`, I was saying that this commit looks OK but I wanted to draw a line for potential future changes that would be more complex than that.

I'm currently preparing the 3.2 branch and I included this commit.

gsmet:
And I just pushed the 3.2 branch!

michalvavrik:
@gsmet I'll keep it in mind. Thanks for pushing it to 3.2. It's part of Windows support testing.

PR #1400, opened 2024-03-21
fix: download file method should have byte[] as response content

petro1:
fix: download file method should have byte[] as response content


scrocquesel:
@gsmet LGTM

gsmet (approved):

petro1:
I see the jdk build failed, what can we do here? Thanks.

PR #1421, opened 2024-04-30
Add OIDC refresh token skew test

sberyozkin:
This PR confirms that the ID token can be refreshed if it is valid but will expire within the configured refresh time skew period. For example, if the ID token is valid for 9 seconds, then it will only live for another 6 seconds after accessing Quarkus after a 3 seconds delay, which falls into a configured 7 seconds skew period.

CC @calvernaz


geoand (approved):

sberyozkin:
optaplanner keeps failing

PR #1457, opened 2024-09-26
Remove reference to nonexistent v1 dev UI from quickstart

holly-cummins:
While I was looking at the index html, I noticed it's out of date in a few other ways (references to reactive, etc) so I've tidied those up as well.


phillip-kruger (commented):
If this is using a later version of Quarkus, you can just remove the index file, as it's dynamic now

holly-cummins:
> If this is using a later version of Quarkus, you can just remove the index file, as it's dynamic now

I realised we do it dynamically after I'd made the changes, but didn't think about the implications for this quickstart. I guess in the development branch it definitely will be a later version of Quarkus, and the less code we have the better, so good idea, I'll remove it. Thanks.

holly-cummins:
Actually, should we remove the index.htmls for *all* quickstarts in this repository?

phillip-kruger:
Yes I think we can remove it from all quickstarts (if they use the latest Quarkus)

holly-cummins:
Closing in favour of #1459.

PR #1459, opened 2024-10-07
Remove template-generated index.htmls

holly-cummins:
See discussion in https://github.com/quarkusio/quarkus-quickstarts/pull/1457. 
New versions of Quarkus automatically generate a landing page, so we do not need an index.html anymore. Since code is a liability, we should reduce it where we can. 

Some quickstarts have an index.html with contents that was hand-crafted, so I left those. The way I distinguished between the two was looking for the text `Congratulations, you have created a new Quarkus application`. This does mean we lose some hand-crafted content on some of the `index.html` titles, but I think we can live with that?


phillip-kruger (approved):

phillip-kruger:
Thanks!

PR #1503, opened 2025-02-17
Update Dockerfiles to Ubi 9

gsmet:
With special attention for the AWT ones.


gsmet:
@Karm I will probably merge it soon as it's in our way but could you please have a look at what I did for the AWT quickstart?

I was able to build the images but not sure they are fine. Also I had to add the installation of `expat` to the micro image and maybe it's not needed anymore and we can drop copying the `.so` instead.

PR #1523, opened 2025-04-14
Add a more complex CRUD example based on the Super Heroes Villain service

cescoffier:


crud-movie-quickstart/src/test/java/org/acme/MoviesEndpointIT.java
```diff
@@ -0,0 +1,171 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static io.restassured.http.ContentType.JSON;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+
+@QuarkusIntegrationTest
+public class MoviesEndpointIT {
```
zakkak, line 14: @cescoffier why is this class duplicating `MoviesEndpointTest`, shouldn't it just be:

```suggestion
public class MoviesEndpointIT extends MoviesEndpointTest {
```

zakkak (commented):
@cescoffier it LGTM but I am a bit concerned about the use of virtual threads. Do we indeed expect this to be the most common scenario?

I also added a note about the duplicated tets.

PR #1543, opened 2025-06-11
Fix Elasticsearch version

gsmet:


yrodiere:
Thanks. cc @marko-bekhta it seems we'll have to remember this, next time :see_no_evil:

PR #1555, opened 2025-07-29
Remove unnecessary comment, workaround is handled by the client

karesti:
I already PR this, but I Pr the wrong branch (main) instead of dev
Docs are updated already


karesti:
@gsmet In case you missed it

Sanne (approved):

PR #1573, opened 2025-10-07
Remove all dependencies from websocket index.html

newur:
**Check list**:

- [ x] targets the `development` branch
- [ x] uses the `999-SNAPSHOT` version of Quarkus
[...]
- [kinda? ] makes sure the associated guide must not be updated
- [ will do if positive feedback] links the guide update pull request (if needed)

====================================

Hello,

I followed the guide on [Getting started with WebSocket Next](https://quarkus.io/guides/websockets-next-tutorial). It all worked fine but I kept wondering about the linked [index.html](https://github.com/quarkusio/quarkus-quickstarts/blob/main/websockets-next-quickstart/src/main/resources/META-INF/resources/index.html). This one HTML file comes with 3 external JavaScript dependencies + 2 CSS dependencies.

Besides the problem that this makes the example less self contained, IMO the file becomes more complicated than it must be. So I would like to propose an alternative - that looks less nice - which does basically the same but with bare bone HTML/JS.

Pro:
- no dependencies that can be outdated
- easier to understand (subjective)

Con:
- looks less fancy
- would require an update of the guide with the less pretty screenshot (if positive feedback here)

Let me know if you think this change is worth it. I could also add "some" CSS to make it less ugly, but honestly from my point of view it should focus on the WebSocket stuff and not on a pleasing UI.

### DIFF
--- a/websockets-next-quickstart/src/main/resources/META-INF/resources/index.html
+++ b/websockets-next-quickstart/src/main/resources/META-INF/resources/index.html
@@ -1,164 +1,173 @@
 <!DOCTYPE html>
-<html>
-
+<html lang="en">
 <head>
     <meta charset="UTF-8">
-    <title>Quarkus Chat!</title>
-    <link rel="stylesheet" type="text/css"
-          href="https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.24.0/css/patternfly.min.css">
-    <link rel="stylesheet" type="text/css"
-          href="https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.24.0/css/patternfly-additions.min.css">
+    <title>Quarkus Chat</title>
 
     <style>
+        body {
+            font-family: Arial, Helvetica, sans-serif;
+        }
+
+        input {
+            padding: 0.3em 1em;
+            background: WhiteSmoke;
+            border: 1px solid lightgray;
+            border-radius: 3px;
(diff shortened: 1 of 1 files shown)


newur:
Thanks for the feedback.

I agree, it does look a bit "rough". Please see my second commit, which adds some CSS to look closer to the original.

Screenshot:
Screenshot from 2026-02-25 20-21-57

cescoffier:
That is great! I will have a proper look, but already thanks for that work!

cescoffier:
Can you squash your commits? Then, it's good to go!

newur:
Thanks for the approval. Yeah, I will squash the commits in the end, but...

Lately I read about `setHTML` (unfortunately [only supported in Firefox](https://hacks.mozilla.org/2026/02/goodbye-innerhtml-hello-sethtml-stronger-xss-protection-in-firefox-148/), yet) and how pairing `innerHTML` and alike with user input is basically an invitation for XSS attacks.

The original version and my initial commit are both vulnerable to inputs like ``. I know this is just a hello world demo, but somehow it feels wrong to re-do the `index.html` and ignore a security problem, once one is aware of it.

So long story short, please review again my latest commit. It is a little bit less "simple", but I would say the additional security is worth it. What do you think @cescoffier ?

newur:
Is it good to merge? @cescoffier

cescoffier (approved):

cescoffier:
Thanks!

PR #1589, opened 2025-11-26
Fix failing stork-kubernetes-quickstart test after Fabric8 7.2.0 upgrade

xstefank:
Required to fix the problem introduced by https://github.com/quarkusio/quarkus/pull/50828.


xstefank:
@jmartisk so the CI isn't testing with the latest 3.20 build it seems.

jmartisk:
The CI runs against 3.20.4, it doesn't build a 3.20 snapshot... And the quickstart will start failing after this PR is merged, but I guess there's nothing else we can do - we don't have a `development` branch for quickstarts that depends on snapshots...

PR #1590, opened 2025-12-01
Switch to Testcontainers 2.x

gsmet:
Draft for now as we will need to merge the PR in main.


amazon-dynamodb-quickstart/pom.xml
```diff
@@ -73,12 +73,12 @@
         </dependency>
         <dependency>
```
scrocquesel: These explicit dependencies on testcontainers should not be necessary with devcontainers. I think it is some leftovers.

gsmet: Makes sense, I dropped them, thanks!

PR #1623, opened 2026-02-27
Update surefire plugin

holly-cummins:
The version of Surefire has got rather out of sync with what's in the Quarkus codebase. It looks better if our quickstarts have up to date dependencies. More importantly, some behaviour differences in Surefire affect how tests run, so we want to use the version of Surefire that the Quarkus test framework was designed for. For example,  https://github.com/quarkusio/quarkus/pull/51298 is a case where running recent versions of Quarkus with Surefire 3.5.3 and lower would leak memory, because of a change in how many `LauncherSession`s Surefire uses.

### DIFF
--- a/amazon-dynamodb-quickstart/pom.xml
+++ b/amazon-dynamodb-quickstart/pom.xml
@@ -11,8 +11,8 @@
 
     <properties>
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-        <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-        <quarkus.platform.version>999-SNAPSHOT</quarkus.platform.version>
+        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+        <quarkus.platform.version>3.32.0.CR1</quarkus.platform.version>
         <surefire-plugin.version>3.1.2</surefire-plugin.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>17</maven.compiler.source>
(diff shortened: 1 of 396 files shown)


gsmet:
You should target the `development` branch.

gsmet:
We can merge this one in `main` for 3.32 and 3.33 though so I'll merge.

gsmet (approved):

gsmet:
Actually, I will merge after we got a CI run in `development` if you don't mind :).

holly-cummins:
I propose we swap round the names and targets, and use 'main' for the branch people are supposed to target, and something like 'upcoming' or 'future' for the branch people might want to target for future-only work. :) 

But I did wonder why CI didn't trigger, so I'll go raise on 'development'.

gsmet:
To our defense, you dropped the pull request template that starts with checking the branch.

It's not as easy to find something that works when this repository is here for people to consume it as examples.

So we made things a bit harder for developers and a bit easier for consumers, which I think is the right decision in this case.

holly-cummins:
> To our defense, you dropped the pull request template that starts with checking the branch.

I did look at the list, but all the other entries weren't relevant for what I was doing, so I assumed the first one was also not relevant. :) 

> It's not as easy to find something that works when this repository is here for people to consume it as examples.
> 
> So we made things a bit harder for developers and a bit easier for consumers, which I think is the right decision in this case.

Yes, I agree that's the right balance. The more opportunities we provide for people to make mistakes, the more mistakes will be made – but the volume of mistake-makers is greater among consumers than among developers, just because there's more of them.  

I wonder if this is a good thing for a bot to catch, to save you pointing it out every time?

holly-cummins:
Current `main` and `development` both look good, so I think this got implemented via another PR. Closing this.

PR #1633, opened 2026-03-30
Stork - Use consistent defaults between Registration and *Services

gsmet:
@aureamunoz not sure what the value should be but we shouldn't have inconsistent defaults for the same `@ConfigProperty`.

Could you have a look at the PR?


aureamunoz (approved):
Yes, you're right. Thank you very much

PR #1635, opened 2026-04-07
Fix blocking issue causing test failures in context-propagation and jta-quickstarts

Andy-Y1314:
### Summary
Fix failing native tests in quickstarts from github action caused by a change in Quarkus, where @'Transactional' methods are no longer implicitly considered blocking.

Fixes https://github.com/quarkusio/quarkus-quickstarts/issues/1636

### Changes
Added @'Blocking' to methods that have @'Transactional' annotation in:
- context-propagation-quickstart
- jta-quickstart

### Reference
https://github.com/quarkusio/quarkus/wiki/Migration-Guide-3.35#quarkus-rest

### Contributors
@SlimaneAbzar

### DIFF
--- a/context-propagation-quickstart/src/main/java/org/acme/context/EmitterResource.java
+++ b/context-propagation-quickstart/src/main/java/org/acme/context/EmitterResource.java
@@ -12,6 +12,7 @@ import jakarta.ws.rs.core.MediaType;
 import org.eclipse.microprofile.reactive.messaging.Channel;
 import org.jboss.resteasy.reactive.RestStreamElementType;
 
+import io.smallrye.common.annotation.Blocking;
 import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.infrastructure.Infrastructure;
 
@@ -23,6 +24,7 @@ public class EmitterResource {
     @Channel("prices") Multi<Double> prices;
 
     @Transactional
+    @Blocking
     @GET
     @Path("/prices")
     @Produces(MediaType.SERVER_SENT_EVENTS)
--- a/jta-quickstart/src/main/java/org/acme/quickstart/TransactionalResource.java
+++ b/jta-quickstart/src/main/java/org/acme/quickstart/TransactionalResource.java
@@ -2,6 +2,8 @@ package org.acme.quickstart;
 
 import com.arjuna.ats.jta.TransactionManager;
 import io.quarkus.runtime.StartupEvent;
+import io.smallrye.common.annotation.Blocking;
+
 import org.eclipse.microprofile.context.ManagedExecutor;
 import org.eclipse.microprofile.context.ThreadContext;
 import org.eclipse.microprofile.context.spi.ContextManagerProvider;
@@ -56,6 +58,7 @@ public class TransactionalResource {
     @POST
     @Produces(MediaType.TEXT_PLAIN)
     @Transactional
+    @Blocking
     @Path("cmt")
     public int cmt() throws SystemException {
         // the @Transactional annotation will have started a transaction
@@ -83,6 +86,7 @@ public class TransactionalResource {
     @POST
     @Produces(MediaType.TEXT_PLAIN)
     @Transactional
+    @Blocking
     @Path("async-with-suspended")
     public void async1(@Suspended AsyncResponse ar) throws SystemException {
         // the framework will have started a transaction because of the @Transactional annotation
@@ -118,6 +122,7 @@ public class TransactionalResource {
     @POST
     @Produces(MediaType.TEXT_PLAIN)
     @Transactional
+    @Blocking
     @Path("async-with-completion-stage")
     public CompletionStage<Integer> async2() throws SystemException {
         System.out.printf("submitting async job ...%n");
@@ -147,6 +152,7 @@ public class TransactionalResource {
 
     @POST
     @Produces(MediaType.TEXT_PLAIN)
+    @Blocking
     @Transactional
     @Path("async-6471-reproducer")
     public void asyncIssue6471Reproducer(@Suspended AsyncResponse ar) throws SystemException {


rsvoboda:
CC @lucamolteni @yrodiere @gsmet 

Quickstarts were not updated with https://github.com/quarkusio/quarkus/pull/51063 change

rsvoboda:
@gastaldi for cases like this we need that GIB thingy in place ;)

rsvoboda:
https://github.com/quarkusio/quarkus/pull/51063 PR updates 
 * docs/src/main/asciidoc/hibernate-reactive-panache.adoc
 * docs/src/main/asciidoc/hibernate-reactive.adoc

But no update to Quarkus REST guides.

So I'm wondering if
 * there shouldn't be updates to Quarkus REST related guides
 * some of the code examples we have across the guides do not need the `@Blocking` addition

Was this checked @lucamolteni as part of your work?

lucamolteni:
> [quarkusio/quarkus#51063](https://github.com/quarkusio/quarkus/pull/51063) PR updates
> 
> * docs/src/main/asciidoc/hibernate-reactive-panache.adoc
> * docs/src/main/asciidoc/hibernate-reactive.adoc
> 
> But no update to Quarkus REST guides.
> 
> So I'm wondering if
> 
> * there shouldn't be updates to Quarkus REST related guides
> * some of the code examples we have across the guides do not need the `@Blocking` addition
> 
> Was this checked @lucamolteni as part of your work?

No sorry I might have missed that 

Anyway the fix is correct

lucamolteni (approved):

yrodiere:
Looking at `jta-quickstart/src/main/java/org/acme/quickstart/TransactionalResource.java`, the changes make perfect sense to me, and match our previous discussions on Zulip: the methods exposes a signature that looks async, but really are blocking so they need to execute on a worker thread and they need a `@Blocking` annotation.

If such use cases are documented, yes we should update the documentation.

---

However, seeing this code... I wonder if we're not allowing too much.

```java
    @Transactional
    @Blocking // NEW
    @GET
    @Path("/prices")
    @Produces(MediaType.SERVER_SENT_EVENTS)
    @RestStreamElementType(MediaType.TEXT_PLAIN)
    public Multi<Double> prices() {
        // get the next three prices from the price stream
        return Multi.createFrom().publisher(prices)
                .select().first(3)
                // The items are received from the event loop, so cannot use Hibernate ORM (classic)
                // Switch to a worker thread, the transaction will be propagated
                .emitOn(Infrastructure.getDefaultExecutor())
                .map(price -> {
                    // store each price before we send them
                    Price priceEntity = new Price();
                    priceEntity.value = price;
                    // here we are all in the same transaction
                    // thanks to context propagation
                    priceEntity.persist();
                    return price;
                    // the transaction is committed once the stream completes
                });
    }
```

The fact that `@Transactional` is set at the method level, and the comments, suggest that the transaction covers the whole stream.
This seems a bit wild in the general case where we're not just retrieving the first 3 items but instead are retrieving many items over a (presumably) long period of time, but okay; let's admit people know what they do and know how to avoid or deal with long-running transactions. I wouldn't want to be them.

The fact we're using `@Blocking`, and that previously `@Transactional` was interpreted as meaning `@Blocking`, tells me this method will be executed on a worker thread. But that's not strictly necessary from an execution perspective: the only blocking code I see is in a callback that we take care of executing in a worker thread.
So we're only executing this stuff on a worker thread so that Narayana can set its thread-local and so that it will be propagated to the rest of the stream through context propagation? Or to put it differently, we create a pipeline from a worker thread so that we can start a blocking transaction on that worker thread and have it propagated to the pipeline, through non-blocking steps and (ultimately) to blocking steps?

Not to mention... this:

```
                    // the transaction is committed once the stream completes
```

Please, please tell me that:

1. a client closing the stream early will trigger a rollback
2. the server closing the stream (because it reached the last item) and encoutering a transaction rollback will be communicated to the client somehow, in the last packet or something...

In the end it still seems @lucamolteni's change is only making existing problems more obvious, but this is all very concerning... Are people really doing this?

@FroMage do you know, or do you know who would know?

rsvoboda:
CC @ozangunalp another RabbitMQ QS fail

FroMage:
@yrodiere we recently made sure that the client closing the connection in Quarkus REST would call `cancel` on `Uni` but I don't remember if it was also for `Multi`.

See https://github.com/quarkusio/quarkus/issues/51576 and there was another one but I can't seem to find it yet.

As for:

> the server closing the stream (because it reached the last item) and encoutering a transaction rollback will be communicated to the client somehow, in the last packet or something

Well, the problem with SSE is that there's no underlying protocol for notifying of any error during the stream, because all HTTP status packets have already been passed in the first response. It's a known issue, but I'm not sure if there's any known solution.

yrodiere:
> @yrodiere we recently made sure that the client closing the connection in Quarkus REST would call `cancel` on `Uni` but I don't remember if it was also for `Multi`.
> 
> See [quarkusio/quarkus#51576](https://github.com/quarkusio/quarkus/issues/51576) and there was another one but I can't seem to find it yet.

That's good but considering the transactional interceptor is on the method, and the method needs to finish executing before we can subscribe to the `Multi` (not to mention cancel the subscription), I doubt that the cancellation will affect the transaction.

In fact I don't even understand how this code could ever lead to a commit; I'd expect the transaction interceptor to trigger the commit when the method returns the `Multi`, so before any item has even been processed...

Yeah, I probably don't understand what this code does. I'd need to run it.

cescoffier:
@yrodiere I would have said it waits for the completion signal or failure. There is an operator that tells you if the streams terminates because of a cancellation (-> rollback), failure (-> rollback), or completion (-> commit)

yrodiere:
> @yrodiere I would have said it waits for the completion signal or failure

You're right, someone added code to the blocking transaction interceptor so that, if the return type is async (and the execution model is blocking, for some reason), commit is delayed until the end of async execution...

Anyone interested, it's there: https://github.com/quarkusio/quarkus/blob/e14cecfcb72a3086d2d3ac02d9dc9dcb5c1061e8/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java#L273-L343

For some odd reason, `Uni` is not supported.

FroMage:
@yrodiere that's because anything not of standard async types is automatically converted:

https://github.com/quarkusio/quarkus/blob/e14cecfcb72a3086d2d3ac02d9dc9dcb5c1061e8/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java#L174

This used to be the pattern for supporting any reactive lib, back before Mutiny was made. I guess we could definitely improve this now.

yrodiere:
> I guess we could definitely improve this now.

I guess... though to be honest, an async pipeline with synchronous, blocking bits where the whole async processed is wrapped in a single blocking transaction... that looks very fragile to me. I suspect anyone using that will have problems down the line.

If I had to change anything, it would be to only support `@Transactional` for code that stay purely reactive or purely blocking, and to recommend *explicit* solutions for anything that goes beyond that.
In the example above that could translate into collecting all prices, and persisting them in one single blocking process on `Multi` completion.
Or that could translate into somehow modeling the fact there are actually two processes happening here: the blocking process that waits for `Price` instances and persists them, and the reactive one that takes an incoming price, turns it into a `Price`, sends it to the blocking process, "waits" for the ID to be assigned, and returns the instance to the client.
Of course this looks like a lot of ceremony, but then the example is contrived; for something so simple I'd just use direct SQL, or Hibernate Reactive, and then only with a stateless session. For something more complex, the ceremony might actually help understand what is actually going on.

rsvoboda:
I'm going to merge this PR so we have less noise in QS CI runs

FroMage:
@yrodiere it is a very weird use-case to extend a transaction over multiple SSE emissions. For a read-only TX it might make some sort of sense, except for long streams. But for a TX that writes stuff on emissions, I hardly see any justification or sanity, especially if the TX drags on until the end of an SSE, as fragile as that is.

yrodiere:
> @yrodiere it is a very weird use-case to extend a transaction over multiple SSE emissions. For a read-only TX it might make some sort of sense, except for long streams. But for a TX that writes stuff on emissions, I hardly see any justification or sanity, especially if the TX drags on until the end of an SSE, as fragile as that is.

Agreed. Hence: if it's insane, let's not support it out of the box :)


EDIT: To clarify, by "explicit solutions", I meant mainly making sure the building blocks are there, and documenting/testing it's possible. As opposed to implicitly supporting `@Transactional` on `Multi`.

FroMage:
You want to remove support for this?

I bet someone will complain. Though we could phase this out with deprecation messages.

yrodiere:
> You want to remove support for this?
> 
> I bet someone will complain. Though we could phase this out with deprecation messages.

Yes, three times :)

TBH we don't need to remove it, we can just hide it behind a flag that's disabled by default and keep it there until the last user has migrated.

EDIT: What we do need is to document the alternative solution.

FroMage:
Open an issue, I guess 🤷

yrodiere:
> Open an issue, I guess 🤷

https://github.com/quarkusio/quarkus/issues/53570

PR #1644, opened 2026-05-06
Fix stork quickstart as it's missing now optional `vertx-codegen`

jedla97:
The https://github.com/quarkusio/quarkus/pull/53924 bumped `smallrye-mutiny-vertx-binding` to 3.22.2 where they make `vertx-codegen` optional. This caused the stork quickstart to fail. This PR adding it

**Check list**:

Your pull request:

### DIFF
--- a/stork-quickstart/pom.xml
+++ b/stork-quickstart/pom.xml
@@ -45,6 +45,10 @@
             <artifactId>smallrye-mutiny-vertx-consul-client</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-codegen</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


cescoffier (changes requested):
I don't believe it's right. The fix should be in stork if stork depends on this. 

/CC @jponge @aureamunoz

jponge:
Agreed, and why are these annotations needed? They should only be used on interfaces for which we want to generate bindings, never for other kinds of code (e.g., `@Fluent` is really a code generator hint, and `@Nullable` should be replaced by a null-safe annotation like JSpecify or Smallrye Commons).

jedla97:
Closing as it done by https://github.com/quarkusio/quarkus-quickstarts/pull/1645

PR #1658, opened 2026-07-22
Fix compatibility with Quarkus 4

gsmet:


gsmet:
@marko-bekhta not sure if the code is in the guide but we might also have to update the Elasticsearch guide if it is.

marko-bekhta:
ohh ... I did try to keep the docs part in sync in that pr:

https://github.com/quarkusio/quarkus/blob/a6b51c11611551beb0a5ebbbbef0b097c6be0369/docs/src/main/asciidoc/elasticsearch.adoc?plain=1#L134-L141

and we have a similar test (though it's a resource not a service):

https://github.com/quarkusio/quarkus/blob/a6b51c11611551beb0a5ebbbbef0b097c6be0369/extensions/elasticsearch-rest-client/deployment/src/test/java/io/quarkus/elasticsearch/restclient/lowlevel/runtime/TestResource.java#L24-L35

And it seems fine throwing the `IOException` 🤔 I wonder how the quickstart is different .... 
le me go check 🏃🏻

marko-bekhta:
yes, looks like it throws an IO one: https://github.com/elastic/elasticsearch-java/blob/bb58d499c679e9025519d2e24b717486a0d57d57/rest5-client/src/main/java/co/elastic/clients/transport/rest5_client/low_level/Rest5Client.java#L291

gsmet:
The problem is with `EntityUtils` throwing also a `ParseException`.

marko-bekhta (approved):

vertx-quickstart/src/main/java/org/acme/extra/StreamingResource.java
```diff
@@ -1,26 +1,23 @@
...
 
-    @Inject
-    Vertx vertx;
-
     @GET
     @Produces(MediaType.SERVER_SENT_EVENTS)
     @Path("{name}/streaming")
     public Multi<String> greeting(String name) {
-        return vertx.periodicStream(2000).toMulti().map(l -> String.format("Hello %s! (%s)%n", name, new Date()));
+        return Multi.createFrom().ticks().every(Duration.ofMillis(2000))
```
cescoffier, line 20: Hum, this should not be required. @jponge, did this api changed?

jponge: Vert.x 5 doesn't have periodic streams anymore, just periodic callbacks, so this is correct. Added bonus: Mutiny ticks are on the event-loop when hooked on Quarkus.

PR #1679, opened 2026-09-10
Fix (workaround) gRPC native tests

alesj:
This should fix the native tests.
But is it a workaround or legit fix?
e.g. should we address this issues in Quarkus core, @cescoffier, @anavarr ?


alesj:
#6588

cescoffier (commented):
I think we need to proper fix in the gRPC extension that switch to TLS (and port) when a certificate is configured.

